PR #210, opened 2018-11-16
Fix: Use Sun - Sat weeks for NALD

swaff:
WATER-1632

Changes week calculations to us weeks that start on Sundays.

Sets the global moment locale to en-gb, but then adds a nald-dates
module for calculating the weeks, where a local locale of `en` is used
which treats weeks as running Sun - Sat.

Updates any existing code that used `startOf` or `endOf` and routes
through nald-dates.

### DIFF
--- a/src/modules/returns/lib/model-returns-mapper.js
+++ b/src/modules/returns/lib/model-returns-mapper.js
@@ -2,6 +2,7 @@ const { get, pick } = require('lodash');
 const moment = require('moment');
 const { convertToCubicMetres, convertToUserUnit } = require('./unit-conversion');
 const uuidv4 = require('uuid/v4');
+const naldDates = require('../../../lib/nald-dates');
 
 /**
  * Converts a line from the returns service to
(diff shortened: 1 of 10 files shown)


src/modules/returns/lib/model-returns-mapper.js
```diff
@@ -92,13 +93,14 @@ const getYears = (startDate, endDate) => {
...
   do {
+    const week = naldDates.getWeek(datePtr);
     lines.push({
-      startDate: datePtr.startOf('week').format('YYYY-MM-DD'),
-      endDate: datePtr.endOf('week').format('YYYY-MM-DD'),
+      startDate: week.start.format('YYYY-MM-DD'),
+      endDate: week.end.format('YYYY-MM-DD'),
       timePeriod: 'week'
     });
     datePtr.add(1, 'week');
```
james-siteclick: Do we maybe need the same fix proposed in the import service?

PR #278, opened 2019-01-25
Feat: Adds user status resource

swaff:
WATER-1833

Adds a new resource at `/user/id/status` that gives an overview of the
licences and companies a user is associated with.


src/modules/users/controller.js
```diff
@@ -0,0 +1,80 @@
...
+  return documentHeaders
+    .filter(doc => doc.company_entity_id === companyEntityId)
+    .map(doc => ({
+      documentId: doc.document_id,
+      licenceRef: doc.system_external_id,
+      licenceHolder: get(doc, 'metadata.contacts[0].name', '')
+    }));
+};
+
+const getCompanyOutstadingVerifications = (company, verifications) => {
```
james-siteclick: Should this be getCompanyOutstandingVerifications?

PR #313, opened 2019-02-15
Adds an API endpoint to review an uploaded returns JSON file

james-siteclick:


src/modules/returns/controller.js
```diff
@@ -120,9 +121,55 @@ const postUploadXml = async (request, h) => {
...
+ */
+const postUploadPreview = async (request, h) => {
+  const { eventId } = request.params;
+  const { companyId } = request.payload;
+
+  try {
+    // Load event - 404 if not found
+    const evt = await Event.load(eventId);
+    if (!evt) {
+      return notFound(h, `Return upload event not found`, eventId);
```
swaff: Could we use Boom here?

PR #331, opened 2019-02-26
Feature/1935 updated json to xml mapping

swaff:


src/lib/connectors/idm.js
```diff
@@ -1,23 +1,23 @@
...
+    external_id: { $in: ids }
   });
 };
 
-module.exports = {
-  usersClient
+/**
+ * Find a single user that has the given user name
+ */
+usersClient.getUserByUserName = async userName => {
```
james-siteclick: Do we need to limit this to the water_vml application in the IDM?

src/modules/returns/lib/xml-to-json-mapping.js
```diff
@@ -41,19 +55,27 @@ const wasMeterUsed = (meterUsage) => {
...
-  const unitOfMeasurement = ret.get('tns:GorPart', options).get('tns:FullReturnStructure', options).get('tns:UnitOfMeasurement', options).text();
-  if (unitOfMeasurement === 'CubicMetres') return 'm³';
-  return null;
+  const unitOfMeasurement = ret
+    .get('tns:GorPart', options)
+    .get('tns:FullReturnStructure', options)
+    .get('tns:UnitOfMeasurement', options)
+    .text();
+
+  return (unitOfMeasurement === 'CubicMetres') ? 'm³' : null;
```
james-siteclick: Are we certain that only CubicMetres is supported in the XML, or do we need to support other units

src/modules/returns/lib/jobs/xml-to-json.js
```diff
@@ -40,9 +41,12 @@ const handleReturnsXmlToJsonStart = async job => {
...
 
   try {
-    const s3Object = await returnsUpload.getReturnsS3Object(job.data.eventId);
+    const [s3Object, userResponse] = await Promise.all([
+      returnsUpload.getReturnsS3Object(job.data.eventId),
+      idmConnector.usersClient.getUserByUserName(evt.issuer)
+    ]);
 
-    const json = mapXml(parseXmlFile(s3Object.Body));
+    const json = await xmlToJsonMapping.mapXml(parseXmlFile(s3Object.Body), userResponse.data[0]);
```
james-siteclick: Do we need to check the user has been loaded correctly before passing to the mapping function?

PR #334, opened 2019-02-27
Feat: Use common return functions from helpers library

swaff:
WATER-1716

Functions to calculate return lines and NALD dates have been moved to
the water-abstraction-helpers library so this change deletes the
implementation in this project and makes use of the extracted functions.


src/modules/pdf-notifications/view-helpers.js
```diff
@@ -2,7 +2,7 @@ const moment = require('moment');
 const { chunk } = require('lodash');
 const naldRegions = require('./data/nald-regions');
 const naldAreas = require('./data/nald-areas');
-const naldDates = require('../../lib/nald-dates');
+const naldDates = require('@envage/water-abstraction-helpers/src/nald/dates');
```
james-siteclick: Is it OK to be requiring things from the /src directory or is it best to go via an object path?

PR #335, opened 2019-02-28
Feat/water 1929

james-siteclick:
Adds endpoints to preview both a batch of uploaded returns and a single return from the batch.


src/modules/returns/controller.js
```diff
@@ -123,24 +128,67 @@ const postUploadXml = async (request, h) => {
...
+
+    if (!match) {
+      throw Boom.notFound(`Return ${returnId} not found in upload`);
+    }
+
+    const validated = await uploadValidator.validate([match], companyId);
+
+    // Load return from return service in order to get metadata
+    const response = await returnsConnector.returns.findOne(returnId);
+    throwIfError(response.error);
```
swaff: I think these to things to happen at the same time.

e.g. something like

```
const [validated, returnsResponse] = await Promise.all([
  uploadValidator.validate([match], companyId),
  returnsConnector.returns.findOne(returnId)
])
```

src/modules/returns/pre-handlers.js
```diff
@@ -42,7 +42,8 @@ const preLoadJson = async (request, h) => {
 const preCheckIssuer = async (request, h) => {
   const { evt } = request;
   const { eventId } = evt;
-  const { userName } = request.payload;
+  const { userName } = request.query;
+  console.log(userName);
```
swaff: ✂️

test/modules/returns/controller.js
```diff
@@ -171,12 +184,87 @@ experiment('postUploadPreview', () => {
...
+  });
+
+  afterEach(async () => {
+    sandbox.restore();
+  });
+
+  test('throws a 404 error if the requested return ID is not in the array', async () => {
+    const request = requestFactory('z');
+    try {
+      await controller.getUploadPreviewReturn(request, h);
```
swaff: I think we should add a `fail` call here in case in the future the error is not thrown as part of a refactor or something. The test will silently pass in that case.

```
try {
  await controller.getUploadPreviewReturn(request, h);
  fail('this should never happen')
}.....
```

PR #338, opened 2019-03-01
Feat: Adds a script to create/delete test returns

swaff:
Usage:

`node scripts/returns create the-return-id`
`node scripts/returns show`
`node scripts/returns delete the-return-id`
`node scripts/returns delete-all`


scripts/returns/data-access.js
```diff
@@ -0,0 +1,124 @@
...
+  returnsConnector.versions.findAll({ return_id: returnId });
+
+/**
+ * Deletes all return lines that have the given version ids.
+ *
+ * @param {Array} asyncversionIds The list of version ids to delete matching lines for
+ * @returns {Promise} A promise that will resolve to all deleted line calls
+ */
+const deleteLines = async versionIds => {
+  // get all the lines for the version ids
```
james-siteclick: Although I don't think it's documented, I think you should be able to delete multiple lines with a single request:
```
returnsConnector.lines.delete({
    version_id: {
      '$in': versionIds
    }
  });
```
I'll add an issue in Github on the hapi-pg-rest-api that delete many is not documented.

PR #394, opened 2019-04-11
Feat: Include expired licences in internal search

swaff:
WATER-1976

Updates the internal search to include expired licences as well when
search for a licence internally.

Allow communications to be acquired for expired licences

Adds partial document details to the licence response
Adds earliest expiry date and reason to the licence response

### DIFF
--- a/src/modules/licences/controller.js
+++ b/src/modules/licences/controller.js
@@ -70,7 +70,7 @@ const wrapData = data => ({
   data
 });
 
-const addEarlistEndDate = licence => {
+const addEarliestEndDate = licence => {
   const dates = [
     { key: 'expired', date: licence.licence_data_value.EXPIRY_DATE },
     { key: 'lapsed', date: licence.licence_data_value.LAPSED_DATE },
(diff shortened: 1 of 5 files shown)


src/modules/licences/controller.js
```diff
@@ -56,18 +70,48 @@ const wrapData = data => ({
   data
 });
 
+const addEarlistEndDate = licence => {
```
james-siteclick: addEarliestEndDate

src/modules/licences/controller.js
```diff
@@ -56,18 +70,48 @@ const wrapData = data => ({
...
+      date.date = moment(date.date, 'DD/MM/YYYY');
+      return date;
+    })
+    .filter(date => date.date.isValid())
+    .sort((l, r) => {
+      if (l.date.isSame(r.date)) return 0;
+      return l.date.isBefore(r.date) ? -1 : 1;
+    })
+    .map(date => {
+      date.date = date.date.format('YYYYMMDD');
```
james-siteclick: Generally the date format in the application is YYYY-MM-DD - just wondering if there's a reason for using YYYYMMDD?

src/modules/licences/controller.js
swaff: Bah, I did that so many other times as well.

PR #420, opened 2019-05-02
Feat: Adds XML upload tests

swaff:
WATER-2117

Adds an integration test that takes example xml returns data, transforms
it to JSON, then validates it using the return model schema to ensure
that these components work together and to provide an early warning if
any future changes introduced a breaking change.

### DIFF
--- a/test/modules/returns/lib/xml-to-json-integration.js
+++ b/test/modules/returns/lib/xml-to-json-integration.js
@@ -1,0 +1,102 @@
+/**
+ * Integration test to ensure the output of the XML to
+ * JSON mapper outputs JSON that validates against the
+ * returns model schema.
+ */
+
+const { expect } = require('code');
+const sinon = require('sinon');
+const sandbox = sinon.createSandbox();
+const {
+  beforeEach,
+  afterEach,
+  experiment,
+  test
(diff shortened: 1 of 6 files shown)


test/modules/returns/lib/xml-to-json-integration.js
```diff
@@ -0,0 +1,102 @@
...
+    stubReturn('03/79/22/1230', '10083015');
+    const parsed = await getParsedTestFile('nil-return');
+    const returns = await mapXml(parsed, getTestUser(), today);
+
+    returns.forEach(ret => {
+      const result = Joi.validate(ret, returnSchema);
+      expect(result.error).to.be.null();
+    });
+  });
+});
```
james-siteclick: Would it be possible to add a test for the case where the XML file includes meter details?

swaff: The single-yearly-return.xml file that is parsed contains meter details, and the estimated-monthly-return.xml file does not contain meter details so this is covered.

Is this enough?

james-siteclick: Yes that's fine thanks :)

PR #432, opened 2019-05-10
Chore: Updates eslint and adds ignore file

swaff:


src/lib/connectors/db.js
```diff
@@ -13,21 +13,7 @@ pool.on('acquire', () => {
 });
 
 function promiseQuery (queryString, params) {
```
james-siteclick: Just wondering if the shape of the response has changed here

swaff: Good spot. Updated that now thanks.

PR #437, opened 2019-05-13
 Feat: Import data into new schema when in progres

swaff:
WATER-2116

Updates the import process so that the 'import' schema is not deleted
until the import has been completed in a new schema.

### DIFF
--- a/src/modules/import/extract.js
+++ b/src/modules/import/extract.js
@@ -19,6 +19,9 @@ const localPath = './temp/';
 const filePath = path.join(localPath, 'nald_dl.zip');
 const finalPath = path.join(localPath, 'NALD');
 
+const SCHEMA_IMPORT = 'import';
+const SCHEMA_TEMP = 'import_temp';
+
 /**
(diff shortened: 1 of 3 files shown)


src/modules/import/extract.js
```diff
@@ -76,31 +79,39 @@ async function getImportFiles () {
...
-  await dbQuery('DROP SCHEMA IF EXISTS "import" CASCADE');
-  await dbQuery('CREATE schema if not exists "import"');
+async function dropAndCreateSchema (schemaName = SCHEMA_IMPORT) {
+  await dropSchema(schemaName);
+  await createSchema(schemaName);
 }
 
+const swapTemporarySchema = async () => {
+  await dropSchema(SCHEMA_IMPORT);
+  await renameSchema(SCHEMA_TEMP, SCHEMA_IMPORT);
```
james-siteclick: Do we want a:
```
await dropSchema(SCHEMA_IMPORT);
```
Once they've been swapped?

PR #498, opened 2019-07-01
Fix: Updates means of determining return frequency

swaff:
WATER-2153

Changes the method of determining the return frequency when uploading
via CSV format using either the number of data rows (12 for a year, > 12
for daily) or the content of the date field (week ending ....)

Then with a good idea of the frequency, and list of accepted date
formats can be used to attempt to mitigate against client programs auto
formatting dates before export to CSV.

### DIFF
--- a/src/modules/returns/lib/csv-adapter/mapper.js
+++ b/src/modules/returns/lib/csv-adapter/mapper.js
@@ -1,11 +1,28 @@
-const { unzip } = require('lodash');
+const { unzip, flatMap } = require('lodash');
 const common = require('../common-mapping');
 const util = require('util');
 const parseCsv = util.promisify(require('csv-parse'));
 const moment = require('moment');
 const DATE_FORMAT = 'YYYY-MM-DD';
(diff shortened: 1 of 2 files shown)


src/modules/returns/lib/csv-adapter/mapper.js
```diff
@@ -4,8 +4,20 @@ const util = require('util');
...
+// preferred format for dates is D MMMM YYYY, but some applications
+// may output the CSV in another format. This list attempts to deal
+// with the unexpected formats without risking potential
+// crossovers between formats such at DD/MM/YYYY and MM/DD/YYYY
+const GDS_DATE_FORMATS = [
+  'D MMMM YYYY',
+  'DD MMMM YYYY',
+  'D MMM YYYY',
+  'DD MMM YYYY',
+  'D-MMMM-YYYY',
```
james-siteclick: I think one I have seen from Excel frequently is:
01-Apr-18
Presumably this is:  DD-MMM-YY

PR #521, opened 2019-07-19
Feat: Change email functionality

DGajewska:
WATER-2206

Migration to add new templates to water.notify_templates table. Create change-email module, controller and routes. Move notify connector from /modules to /lib & updated relevant file references.


src/modules/change-email/controller.js
```diff
@@ -0,0 +1,130 @@
...
+  } catch (error) {
+    logger.error('Email change error', error);
+    if (error.name === 'EmailChangeError') return error;
+  }
+};
+
+/**
+ * call IDM to check user credentials and create email change record
+ */
+const createEmailChangeRecordIDM = (userId, password) => {
```
james-siteclick: The code that does the API calls should probably live in lib/connectors/idm

src/modules/change-email/controller.js
```diff
@@ -0,0 +1,130 @@
...
+ */
+const createEventObject = (userName, entityId, newEmail, userId) => {
+  return event.create({
+    event_id: uuid(),
+    type: 'user-account',
+    subtype: 'email-change',
+    issuer: userName,
+    entities: [entityId],
+    metadata: {
+      'old-email': userName,
```
james-siteclick: A very minor point, but the metadata would be easier to deal with if the keys are camelcase rather than snake case as they won't need quoting then

src/modules/change-email/controller.js
```diff
@@ -0,0 +1,130 @@
...
+      'new-email': newEmail,
+      'user-id': userId
+    }
+  });
+};
+
+/**
+ * update CRM db row for entity with new email for entity_nm
+ */
+const updateEmailInCRM = async (entityId, newEmail) => {
```
james-siteclick: As above for IDM connector code, this should probably live in lib/connectors/crm

src/lib/connectors/idm.js
```diff
@@ -35,6 +36,35 @@ usersClient.getUserByUserName = async userName => {
   return head(data);
 };
 
+/**
+ * call IDM to check user credentials and create email change record
+ */
+const createEmailChangeRecord = (userId, password) => {
+  const url = `${process.env.IDM_URI}/user/change-email-address/start`;
```
james-siteclick: For all the URLs can we use:
```
config.services.idm 
config.services.crm
```
etc.
Rather than using the environment variables directly

src/modules/change-email/controller.js
```diff
@@ -0,0 +1,68 @@
...
+const scheduledNotifications = require('../../controllers/notifications');
+const changeEmailHelpers = require('./lib/helpers');
+
+/**
+ * call IDM to check authentication details, return result
+ * @param  {Object}  request
+ * @param  {Object}  h
+ */
+const postStartEmailAddressChange = async (request, h) => {
+  const { userId } = request.defra;
```
james-siteclick: I don't think `request.defra` will exist in the water service - this is only present in the UI project where the user is authenticated.
I think we would need to pass the user ID in the request.payload in the same way as you have done for the password.

src/modules/change-email/controller.js
```diff
@@ -0,0 +1,68 @@
...
+};
+
+/**
+ * check security code, log event, send confirmation email to old email address
+ * @param  {Object}  request
+ * @param  {Object}  h
+ */
+const postChangeEmailAddress = async (request, h) => {
+  const { securityCode, entityId } = request.payload;
+  const { userId, userName } = request.defra;
```
james-siteclick: I think these also need to be passed in `request.payload`

src/modules/change-email/controller.js
```diff
@@ -0,0 +1,68 @@
...
+const postGenerateSecurityCode = async (request, h) => {
+  const { verificationId, newEmail } = request.payload;
+  try {
+    const { data: { verificationCode } } = await idm.addNewEmailToEmailChangeRecord(verificationId, newEmail);
+
+    const scheduledNotification = changeEmailHelpers.createNotificationData(
+      'email_change_verification_code_email',
+      newEmail,
+      { verification_code: verificationCode });
+    await scheduledNotifications.repository.create(scheduledNotification);
```
james-siteclick: Does this need to return the IDM response if successful?

src/modules/change-email/controller.js
```diff
@@ -0,0 +1,68 @@
...
+  const { userId, userName } = request.defra;
+
+  try {
+    const { data: { newEmail } } = await idm.verifySecurityCode(userId, securityCode);
+
+    await crm.updateEntityEmail(entityId, newEmail);
+
+    const evt = changeEmailHelpers.createEventObject(userName, entityId, newEmail, userId);
+
+    await event.repo.create(evt);
```
james-siteclick: Does this need to return the IDM response if successful?

src/modules/change-email/lib/helpers.js
```diff
@@ -0,0 +1,50 @@
...
+const createNotificationData = (messageRef, recipient, personalisation) => {
+  return {
+    id: uuid(),
+    recipient,
+    message_type: 'email',
+    message_ref: messageRef,
+    personalisation
+  };
+};
+
```
james-siteclick: I wonder if it may be neat to use the same pattern you made for `sendEmailAddressInUseNotification` for both of  of the message types.
It has a simple API and hides Notify implementation details (e.g. the personalisation object) from the controller.

src/modules/change-email/routes.js
```diff
@@ -0,0 +1,43 @@
+const controller = require('./controller');
+const Joi = require('joi');
+
```
james-siteclick: I think we will need a few extra items (e.g. userId)  in the payload for several of these routes as mentioned in the controller above.

PR #551, opened 2019-08-15
Create tables for charge versions/elements

james-siteclick:


migrations/sqls/20190814154806-charge-versions-up.sql
```diff
@@ -0,0 +1,58 @@
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
+
+CREATE TYPE water.charge_scheme AS ENUM ('alcs', 'sroc');
```
DGajewska: You've got a typo, the current scheme is `acls`

DGajewska: You've got a typo, the current scheme is `acls`

PR #558, opened 2019-08-20
Update internal user check

DGajewska:
Updated internal user checks to use application in config in users controller


src/modules/users/controller.js
```diff
@@ -27,7 +27,7 @@ const getCallingUser = async callingUserId => {
 const mapUserStatus = user => {
   return {
     isLocked: parseInt(user.reset_required) === 1,
-    isInternal: !get(user, 'role.scopes', []).includes('external'),
+    isInternal: user.application === config.idm.application.internalUser,
```
james-siteclick: Could we use the same function defined below to avoid repetition:
```
isInternal : isInternalUser(user),
...
```

PR #569, opened 2019-08-28
Feat: Unlink Licence

DGajewska:
WATER-2237

Add CRM connector to unlink licence. Create unlink-licence route and controller.

### DIFF
--- a/src/modules/unlink-licence/controller.js
+++ b/src/modules/unlink-licence/controller.js
@@ -21,12 +21,18 @@ const getCallingUser = async callingUserId => {
   return user;
 };
 
-const createUnlinkLicenceEvent = (callingUser, document) => {
+const unlinkLicenceInCRM = documentId => {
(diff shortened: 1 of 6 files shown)


src/modules/unlink-licence/controller.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param  {Object} h       - HAPI reply interface
+ * @return {Promise}         [description]
+ */
+const patchUnlinkLicence = async (request, h) => {
+  const { callingUserId } = request.payload;
+  const { documentId } = request.params;
+
+  try {
+    const callingUser = await getCallingUser(callingUserId);
+    const document = await crmDocumentsConnector.unlinkLicence(documentId);
```
james-siteclick: I don't think we check here that the rowCount is 1 and therefore that the record has been modified.
Perhaps the `unlinkLicence` function could do this and return a boolean?
If the record is not modified we could then skip the event log and return a different code.

PR #600, opened 2019-09-20
Removes old returns reminder notification, adds one with same logic a…

james-siteclick:
…s new invitation


james-siteclick:
This is awaiting Notify message IDs

src/modules/batch-notifications/config/returns/index.js
```diff
@@ -0,0 +1,21 @@
...
+};
+
+module.exports = [{
+  prefix: 'RINV-',
+  name: 'Returns: invitation',
+  messageType: 'returnInvitation',
+  schema,
+  getRecipients
+},
+module.exports = {
```
swaff: I think we can remove the second `module.exports` here.

PR #617, opened 2019-10-03
Feat: Adds acceptance test setup and tear down resources

swaff:
WATER-2230

Adds routes at /acceptance-tests to create test data for the acceptance
tests

### DIFF
--- a/src/modules/acceptance-tests/lib/entities.js
+++ b/src/modules/acceptance-tests/lib/entities.js
@@ -1,0 +1,41 @@
+const db = require('../../../lib/connectors/db');
+const entityConnector = require('../../../lib/connectors/crm/entities');
+const { ACCEPTANCE_TEST_SOURCE, TEST_COMPANY_NAME } = require('./constants');
+
+const createIndividual = email => createEntity(email, 'individual');
+
+const createCompany = () => createEntity(TEST_COMPANY_NAME, 'company');
+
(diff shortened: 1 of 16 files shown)


src/modules/acceptance-tests/lib/entities.js
```diff
@@ -0,0 +1,41 @@
...
+  delete from
+  crm.entity_roles
+  where created_by = '${ACCEPTANCE_TEST_SOURCE}';`
+);
+
+const deleteEntities = () => db.query(`
+  delete from
+  crm.entity
+  where source = '${ACCEPTANCE_TEST_SOURCE}';`
+);
```
james-siteclick: Just wondering why the creation of records tends to use the API whereas the deletions call the DB directly?

src/modules/acceptance-tests/routes.js
```diff
@@ -0,0 +1,27 @@
...
+    method: 'POST',
+    path: `/water/${version}/acceptance-tests/tear-down`,
+    handler: controller.postTearDown,
+    config: {
+      description: 'Deletes an data created for use with acceptance tests'
+    }
+  }
+};
+
+if (!isProduction) {
```
james-siteclick: I wonder if:
```
(isTestMode)
// or
(isTestMode || isPreProduction)
```
Would be safest here depending where the test suite needs to run - as we had issues in the past with blacklisting environments rather than whitelisting for test flags.

PR #634, opened 2019-11-01
Feat: Adds new tables for billing

swaff:
WATER-2381

Adds billing tables and updates charge_* tables to use uuid data type


migrations/sqls/20191101101124-billing-tables-up.sql
```diff
@@ -0,0 +1,156 @@
...
+
+create table water.billing_invoice_licences (
+  billing_invoice_licence_id uuid primary key default gen_random_uuid(),
+  billing_invoice_id uuid not null
+    constraint billing_invoice_licences_billing_invoice_id_fkey
+    references water.billing_invoices (billing_invoice_id),
+  company_id uuid not null,
+  contact_id uuid not null,
+  address_id uuid not null,
+  licence_ref varchar not null,
```
Stephan-Hough: should licence_ref also have a foreign key constraint?

swaff: I didn't think so really, because licence_ref would map to a table outside of the current schema.

swaff: So not required due to SOA.

PR #648, opened 2019-11-08
Feat: Find charge versions for supplementary billing

swaff:
WATER-2383

- Adds the water.licences table
- Changes the use of `licence_ref` to `licence_id` in the billing tables
- Creates batch charge versions by looking for licences that require
supplementary billing


src/modules/billing/controller.js
```diff
@@ -41,7 +43,7 @@ const postCreateBatch = async (request, h) => {
 
   // add a new job to the queue so that the batch can be filled
   // with charge versions
-  populateBillingBatchJob.publish(batchEvent.event_id);
+  populateBatchChargeVersionsJob.publish(batchEvent.event_id);
```
james-siteclick: We have some business logic around not allowing a user to begin a billing run of the same type in the same region if there is already one in progress.  
Should we implement this in the controller and respond with `409` code possibly?

PR #651, opened 2019-11-14
Adds PG boss to hapi as a plugin, upgrades PG boss to 3.x

james-siteclick:
A hapi plugin has been created to attach PG boss to hapi in the `water-abstraction-helpers` project.
However lots of code in this module relies on the singleton instance of the message queue.
To allow a transition, a temporary plugin has been created here with the same API as the one in `water-abstraction-helpers`, but which still has the singleton instance of the boss.
This should allow us to rewrite the code over time and eventually use the plugin in `water-abstraction-helpers` version when everything is replaces.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "test": "DATABASE_URL=postgres://postgres:postgres@localhost:5432/permits-test && node scripts/create-gen_random_uuid.js && lab",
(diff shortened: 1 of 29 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "url": "https://github.com/DEFRA/water-abstraction-service"
   },
   "scripts": {
-    "test": "DATABASE_URL=postgres://postgres:postgres@localhost:5432/permits-test lab",
+    "test": "DATABASE_URL=postgres://postgres:postgres@localhost:5432/permits-test && node scripts/create-gen_random_uuid.js && lab",
```
swaff: I don't think we need this line anymore.

PR #652, opened 2019-11-15
Modify billing code to use new PG boss plugin

james-siteclick:
Modifies existing billing code to use the instance of PG boss on server.messageQueue or request.messageQueue.


src/modules/billing/plugin.js
```diff
@@ -0,0 +1,8 @@
+const populateBillingBatch = require('./jobs/populate-billing-batch');
+
+module.exports = {
```
swaff: I wonder if this plugin should call `registerSubscribers` from the deleted `jobs/init` file.

With all the publishing and subscribing that we will need to handle for billing there may end up being quite a lot of logic required which might not obviosuly live in the plugin.

src/modules/billing/controller.js
```diff
@@ -41,7 +41,8 @@ const postCreateBatch = async (request, h) => {
 
   // add a new job to the queue so that the batch can be filled
   // with charge versions
-  populateBillingBatchJob.publish(batchEvent.event_id);
+  const message = populateBillingBatchJob.createMessage(batchEvent.event_id);
+  await request.messageQueue.publish(message);
```
swaff: I don't think we need to wait for this to resolve here.

e.g.

`request.messageQueue.publish(message);` over `await request.messageQueue.publish(message);`

james-siteclick: Without the await, I think it might be possible for the publish() method to throw an error but the controller still return 202

swaff: Good point!

PR #659, opened 2019-11-18
Feat: Populates charge version years

swaff:
WATER-2404

 - Adds water.billing_batch_charge_version_year table
 - Creates entries for charge version years for each financial year in
the range defined in the batch to handle annual/supplementary billing
 - Adds new repository functions to update the batch status


package.json
```diff
@@ -53,15 +53,15 @@
...
+    "commander": "^2.20.3",
     "csv-parse": "^2.4.0",
     "deep-map": "^1.5.0",
     "dotenv": "^4.0.0",
     "es6-error": "^4.1.1",
     "firstline": "^2.0.2",
     "good": "^8.1.2",
     "good-winston": "^4.0.0",
-    "handlebars": "^4.5.2",
+    "handlebars": "^4.5.3",
```
james-siteclick: Is handlebars actually used in the project?  I had a quick look and could only see this reference to it
Likewise `handlebars-helper-equal`

PR #675, opened 2019-11-28
Feat/water 2417 write billing invoices

james-siteclick:
Implements an initial version of the charge processor, including handling of:
- Multiple CRM documents per licence 
- Splits in licence holder/invoice account per document
- Time limited elements 
Does not yet handle:
- Licence agreements
- NALD issues in calculating authorised days when abs period changes mid-year 

The output is an array.

Processes these charges in the `billing.charge_version_process` job, and then maps these to an object graph.  

Invoices within the batch are then persisted to the DB.


src/modules/billing/service/charge-processor/mappers.js
```diff
@@ -0,0 +1,4 @@
+const { mapKeys, isObject } = require('lodash');
+const camelCase = require('camelcase');
+
+exports.camelCaseKeys = obj => isObject(obj) && mapKeys(obj, (value, key) => camelCase(key));
```
swaff: Could this be removed and the code at src/lib/camel-case-keys.js be used?

src/modules/billing/service/charge-processor/model-mapper.js
```diff
@@ -0,0 +1,73 @@
...
+ * @param {Object} data - address data from CRM
+ * @return {Address}
+ */
+const mapAddress = data => {
+  const address = new Address();
+  address.id = data.addressId;
+  address.addressLine1 = data.address1;
+  address.addressLine2 = data.address2;
+  address.addressLine3 = data.address3;
+  address.addressLine4 = data.address4;
```
swaff: Is there a danger here that `data.address4` might be undefined and cause a Joi validation error which requires a string or null?

src/modules/billing/service/charge-version-year.js
```diff
@@ -0,0 +1,64 @@
...
+    financial_year_ending: financialYearEnding,
+    billing_batch_id: batchId
+  } = chargeVersionYear;
+
+  // Process charge data
+  const { error, data } = await chargeProcessor.processCharges(financialYearEnding, chargeVersionId);
+
+  if (error) {
+    const msg = 'Error processing charge version year';
+    logger.error(msg, chargeVersionYear);
```
swaff: `logger.error` would take the error as the second arg. So this would be:

`logger.error(msg, error, { chargeVersionYear });`

src/modules/billing/service/charge-processor/mappers.js
james-siteclick: This should be fixed

src/modules/billing/service/charge-processor/model-mapper.js
james-siteclick: The data here comes from the CRM API call so will always be a string or null

src/modules/billing/service/charge-version-year.js
james-siteclick: `error` here is a string so I re-jigged this to create an error object and send this to logger.error.
Should be fixed

PR #691, opened 2019-12-12
Feat: Adds transactions to the invoice API

swaff:
WATER-2414

Provides routes at

`/water/1.0/billing/batches/{batchId}/invoices/{invoiceId}`

and

`/water/1.0/billing/batches/{batchId}/invoices`

that get invoice data with transactions via the charge module.

Because the charge-module is not live at the moment the API is tolerant
of the requirement for the VPN access and returns an empty array of
transactions if none are available.

Updates the models to include `toJSON` to serialize the desired
properties.

Updates the Invoice model to provide a function that will get a totals
object.


src/lib/models/licence.js
```diff
@@ -36,6 +40,22 @@ class Licence {
...
+  set accountNumber (accountNumber) {
+    Joi.assert(accountNumber, VALID_ACCOUNT_NUMBER);
+    this._accountNumber = accountNumber;
+  }
+
+  get accountNumber () {
+    return this._accountNumber;
+  }
+
+  toJSON () {
```
james-siteclick: As above

src/lib/models/licence.js
```diff
@@ -36,6 +40,22 @@ class Licence {
   get licenceNumber () {
     return this._licenceNumber;
   }
+
+  set accountNumber (accountNumber) {
```
james-siteclick: Do we need this when we already have the InvoiceLicence and InvoiceAccount models

src/lib/models/transaction.js
```diff
@@ -0,0 +1,46 @@
...
+  get isCredit () {
+    return this._isCredit;
+  }
+
+  set isCredit (isCredit) {
+    Joi.assert(isCredit, Joi.boolean());
+    this._isCredit = isCredit;
+  }
+
+  toJSON () {
```
james-siteclick: As above

src/modules/billing/services/transactions-service.js
```diff
@@ -0,0 +1,42 @@
...
+const chargeModuleTransactionsConnector = require('../../../lib/connectors/charge-module/transactions');
+const Transaction = require('../../../lib/models/transaction');
+const { logger } = require('../../../logger');
+
+const mapTransaction = chargeModuleTransaction => {
+  const transaction = new Transaction();
+  transaction.id = chargeModuleTransaction.id;
+  transaction.licenceNumber = chargeModuleTransaction.licenceNumber;
+  transaction.accountNumber = chargeModuleTransaction.customerReference;
+  return transaction;
```
james-siteclick: Is it OK to assign properties (licenceNumber, accountNumber) to a Transaction when they are not usually part of that model?
Perhaps we could keep that data separate but still linked to the transaction so they can still be matched up later
```
return {
  licenceNumber : chargeModuleTransaction.licenceNumber,
  accountNumber : chargeModuleTransaction.customerReference,
  transaction : new Transaction(chargeModuleTransaction.id)
}
```
Also I had assumed the `.id` property of each model is its ID in our local database, whereas here it is being used to hold the external ID.
Do we need to add a `chargeModuleId` field or similar to the Transaction model for this?

src/lib/models/batch.js
```diff
@@ -4,33 +4,18 @@ const FinancialYear = require('./financial-year');
...
-const VALID_GUID = Joi.string().guid().required();
 const VALID_BATCH_TYPE = Joi.string().valid('annual', 'supplementary', 'two_part_tariff').required();
 const VALID_SEASON = Joi.string().valid('summer', 'winter', 'all year').required();
 const VALID_STATUS = Joi.string().valid('processing', 'complete', 'error').required();
 
-class Batch {
+const Model = require('./model');
+
+class Batch extends Model {
   constructor () {
```
james-siteclick: To follow the pattern of the other models which don't have a constructor, do we need to pass the ID to the superclass:
```
constructor(id) {
   super(id);
} 
```
or
```
constructor(...args) {
   super(...args);
} 
```

src/lib/models/invoice.js
```diff
@@ -63,17 +75,82 @@ class Invoice {
...
+  /**
+   * Gets the number of pence that are required to add to the invoice
+   * to satisfy the minimum charge of £25.
+   */
+  getMinimumPaymentTopUp () {
+    const { totalValue } = this.getTotals();
+    return totalValue < MINIMUM_PAYMENT ? MINIMUM_PAYMENT - totalValue : 0;
+  }
+
+  getInvoiceLicenceContacts () {
```
james-siteclick: What are these contacts being used for?  If it's for this view:
https://wrls-prototype.herokuapp.com/bd/charges/supplementary-charges
I think we should be using the invoice account holder company.
Also the invoice licence `contact` is not populated where the CRM `company` is a person - not sure if this has an impact on the code below.

src/modules/billing/controller.js
```diff
@@ -75,23 +78,23 @@ const getBatch = async request => {
...
     ? envelope(invoices, true)
     : Boom.notFound(`No invoices found for batch with id: ${batchId}`);
 };
 
-const getInvoiceDetail = async request => {
-  const { invoiceId } = request.params;
-  const invoice = await repos.billingInvoices.getInvoiceDetail(invoiceId);
+const getBatchInvoiceDetail = async request => {
+  const { batchId, invoiceId } = request.params;
+  const invoice = await invoiceService.getInvoiceForBatch(batchId, invoiceId);
```
james-siteclick: In some other recent work (I think creating/modifying user permissions) we had access/permissions checks at the water service layer.
Should we do similar for all billing routes?

swaff: We don't for the creation of a batch at the moment. But we should do this to this and other routes as part of a story to make better use of the JWT token I think.

james-siteclick: Sounds good

PR #700, opened 2019-12-18
Remove db.query connector which swallowed errors

james-siteclick:
Removes the .query() function in the DB connector.
This swallows any errors thrown.
Replaces with pool.query() instead.


src/modules/regions/controller.js
```diff
@@ -7,7 +7,7 @@ const { throwIfError } = require('@envage/hapi-pg-rest-api');
  * for easier consumption upstream
  */
 const getRegions = async () => {
-  const { data: regions, error } = await regionsConnector.getRegions();
+  const { rows: regions, error } = await regionsConnector.getRegions();
   throwIfError(error);
```
swaff: I _think_ we won't need `throwIfError` now because an error will have already be thrown.

PR #715, opened 2019-12-30
Remove data split on licence holder change.

DGajewska:
WATER-2450

Move company, address and contact data in Invoice Licence into licence holder role in Invoice Licence.


james-siteclick:
Decided to keep existing implementation

PR #724, opened 2020-01-09
Feat: Capture region data for licences

swaff:
WATER-2463

Updates the 'water.licences' table
- Adds the regions field to maintain the historic and charge regions required for charging
- Adds a flag to determine if the licence is owned by a water undertaker
- Adds a unique constraint so that a postgres upsert can be run by the import module when persisting a licence.

### DIFF
--- a/migrations/sqls/20200108170348-add-undertaker-and-regions-to-licences-up.sql
+++ b/migrations/sqls/20200108170348-add-undertaker-and-regions-to-licences-up.sql
@@ -1,0 +1,11 @@
+-- no live task will currently populate licences
+-- so this is required for developers who may have
+-- licence data for local testing.
+truncate water.licences cascade;
+
+alter table water.licences
+  add column is_water_undertaker boolean not null,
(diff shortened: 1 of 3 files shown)


migrations/sqls/20200108170348-add-undertaker-and-regions-to-licences-up.sql
```diff
@@ -0,0 +1,11 @@
...
+-- so this is required for developers who may have
+-- licence data for local testing.
+truncate water.licences cascade;
+
+alter table water.licences
+  add column is_water_undertaker boolean not null,
+  add column regions jsonb not null,
+  add column date_created timestamp not null default now(),
+  add column date_updated timestamp not null default now(),
+  add constraint c_licences_lic_ref_region unique (region_id, licence_ref);
```
james-siteclick: Licence numbers should be unique irrespective of the region (I have confirmed this in NALD data), so I think we just  need a unique constraint on the licence_ref column

PR #727, opened 2020-01-10
Add mapper to generate CM data from service models

james-siteclick:
- Partially implements mappers to map the service models to a charge module charge
- Implements a connector to the charge module to create a transaction
- records the charge module transaction ID in the water.billing_transactions table
- Implements the create-charge billing job

### DIFF
--- a/config.js
+++ b/config.js
@@ -16,6 +16,10 @@ module.exports = {
     }
   },
 
+  billing: {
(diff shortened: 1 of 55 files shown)


james-siteclick:
Some of the implementation will change when we add Bookshelf.js, however it seems worth merging this as other work depends on the CM external_id being populated

PR #731, opened 2020-01-13
Feat: Remove customer from a bill run

swaff:
WATER-2411

- Adds new DELETE route for /batches/{batchId}/account/{accountId}

Checks if the batch exists and if it has invoices for the account, then
deletes the transactions from `water.billing_transactions`.

Requires more work in the future to delete the transactions from the
charge module.

### DIFF
--- a/src/modules/billing/controller.js
+++ b/src/modules/billing/controller.js
@@ -94,7 +94,50 @@ const getBatchInvoiceDetail = async request => {
     : Boom.notFound(`No invoice found with id: ${invoiceId} in batch with id: ${batchId}`);
 };
 
+const deleteAccountFromBatch = async request => {
+  const { batchId, accountId } = request.params;
+  const batch = await repos.billingBatches.getById(batchId);
+
+  if (!batch) {
+    return Boom.notFound(`No batch found with id: ${batchId}`);
+  }
+
+  if (batch.status !== 'review') {
+    return Boom.forbidden(`Cannot delete account from batch (${batchId}) when status is ${batch.status}`);
+  }
+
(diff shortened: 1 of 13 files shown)


src/modules/billing/controller.js
```diff
@@ -94,7 +94,50 @@ const getBatchInvoiceDetail = async request => {
...
+    return Boom.forbidden(`Cannot delete account from batch (${batchId}) when status is ${batch.status}`);
+  }
+
+  const invoices = await invoiceService.getInvoicesForBatch(batchId);
+
+  const invoicesForAccount = invoices.filter(invoice => invoice.invoiceAccount.id === accountId);
+
+  if (invoicesForAccount.length === 0) {
+    return Boom.notFound(`No invoices for account (${accountId}) in batch (${batchId})`);
+  }
```
Stephan-Hough: Are lines 109 to 115 required? Is it not possible to only call the database to delete the transactions and then use the rowCount at line 133 to return the appropriate response i.e. 
if rowCount == 0  
  return Boom.notFound(`No invoices for account (${accountId}) in batch (${batchId})`)
else
  return { transactionsDeleted: rowCount };

swaff: It depends on the outcome of the conversations to have the bulk deletions with the charge module. If we have to delete them individually then the current code gives us reference to the transactions without writing more code. These transactions should have the charge module transaction id, which we will need to pass across.

Possibly a question for our future selves.

Stephan-Hough: OK thanks for explaining.

PR #735, opened 2020-01-15
Feat: Adds endpoint to get batches

swaff:
WATER-2447

Part of the UI integration story, this allows the batches to be
requested by the UI.

### DIFF
--- a/src/lib/models/batch.js
+++ b/src/lib/models/batch.js
@@ -105,7 +105,7 @@ class Batch extends Model {
   }
 
   set dateCreated (value) {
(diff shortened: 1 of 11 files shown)


src/lib/models/batch.js
```diff
@@ -104,6 +104,14 @@ class Batch extends Model {
     return this._status;
   }
 
+  set dateCreated (value) {
+    this._dateCreated = value;
```
james-siteclick: Should we assert this to be a YYYY-MM-DD sting?

PR #741, opened 2020-01-17
Feat/water 2475 agreements to cm

james-siteclick:
Stores section 126/127/130 agreements in billing_transactions table and sends relevant flags to charge module.


james-siteclick:
PR 727 should be merged first, then I will rebase this one

src/modules/billing/services/transactions-service.js
```diff
@@ -113,6 +114,32 @@ const mapChargeToTransactions = (chargeLine, batch) => {
...
+    if (agreement.code === 'S126') {
+      acc.section_126_factor = agreement.factor;
+    }
+    if (agreement.code.startsWith('S130')) {
+      acc.section_130_agreement = agreement.code;
+    }
+    return acc;
+  }, skeleton);
+};
+
```
swaff: I think it would be nice to add functions to the `Agreement` classes for:

 - `isTwoPartTariff()`
 - `isAbatement()`
 - `isCanalAndRiversTrust()`

so we don't have to deal with the codes as much

e.g.

```

const mapAgreementsToDB = agreements => {
  const twoPartTariffAgreement = agreements.find(agreement => agreement.isTwoPartTariff());
  const abatementAgreement = agreements.find(agreement => agreement.isAbatement());
  const canalAndRiversTrustAgreement = agreements.find(agreement => agreement.isCanalAndRiversTrust());

  return {
    section_127_agreement: !!twoPartTariffAgreement,
    section_126_factor: abatementAgreement ? abatementAgreement.factor : 0,
    section_130_agreement: canalAndRiversTrustAgreement ? canalAndRiversTrustAgreement.code : null
  }
};
```

james-siteclick: Thanks - I have made this change.
I also made another subtle change - the section_126_factor is now stored as null rather than 1 if no section 126 agreement is present.
Although the factor is still sent to the CM as 1, I thought this was clearer.

PR #750, opened 2020-01-23
Feat/water 2474 orm

james-siteclick:
Begins to implement repository modules which use Bookshelf.js to fetch related data from the DB.
Moves all mapping code from modules/billing/services to modules/billing/mappers and makes naming convention consistent.
Uses new repository/mappers in 'create charge' job and getBatches controller.


swaff:
Looks great. 

Interested to see if we can remove `install` from the `package.json`

Also work adding the bookshelf dir to the `codeclimate.json` file to remove the duplication errors.

PR #753, opened 2020-01-27
Feat: Delete a batch

swaff:
WATER-2410

Adds new DELETE endpoint that allows a batch to be completed removed
from the charge module (pending implementation) and the local water
schema.

After deletion the event is updated with a status of deleted.


src/modules/billing/services/batch-service.js
```diff
@@ -46,6 +48,22 @@ const getBatches = async (page = 1, perPage = Number.MAX_SAFE_INTEGER) => {
...
 };
 
+const deleteBatch = async batchId => {
+  // TODO: This function is a stub and will eventually
+  // have a defined signature and response behaviour.
+  // If this fails, the batch should not be deleted from
+  // the WRLS database tables.
+  await chargeModuleBatchConnector.deleteBatch();
+
+  await repos.billingBatchChargeVersionYears.deleteByBatchId(batchId);
```
james-siteclick: Is deleting from each table preferable to setting up 'on delete cascade' and then just deleting the batch?

src/modules/billing/controller.js
```diff
@@ -142,9 +127,25 @@ const deleteAccountFromBatch = async request => {
   };
 };
 
+const deleteBatch = async (request, h) => {
+  const { batch } = request.pre;
+  const batchId = batch.billing_batch_id;
+
+  await batchService.deleteBatch(batchId);
+
+  const evt = await eventService.getEventForBatch(batchId);
```
james-siteclick: We discussed earlier writing a new event here rather than updating the existing one

src/modules/billing/services/event-service.js
```diff
@@ -0,0 +1,14 @@
+const { get } = require('lodash');
+const event = require('../../../lib/event');
+
+const getEventForBatch = async batchId => {
```
james-siteclick: This wouldn't be needed if we made a fresh event

src/modules/billing/services/batch-service.js
swaff: I prefer it because it is explicit and testable.

PR #766, opened 2020-02-04
Feat: Create transaction key on save

swaff:
WATER-2467

Creates a hash of a transaction when saving a transaction so it can be
more easily compared to other transactions later on.


src/lib/models/transaction.js
```diff
@@ -185,9 +178,60 @@ class Transaction extends Model {
...
+      authorisedDays: this.authorisedDays,
+      volume: this.volume,
+      agreements: this.agreements.map(ag => ag.code).sort().join('-'),
+      accountNumber,
+      source: this.chargeElement.source,
+      season: this.chargeElement.season,
+      loss: this.chargeElement.loss,
+      description: this.description,
+      licenceNumber: licence.licenceNumber,
+      regionCode: licence.region.code
```
james-siteclick: Comparing this with what is sent to the CM, is it still missing a few properties, such as:
* type (compensation|standard)
* twoPartTariff (this is taken from the containing batch and is distinct from whether a TPT agreement applies).

PR #773, opened 2020-02-07
Feat: Bookshelf implementation for Events

Stephan-Hough:
Feat/water 2474


migrations/sqls/20200205114716-event-id-default-uuid-down.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE water.events ALTER COLUMN event_id SET DEFAULT NULL;
```
DGajewska: This is column currently is set to have a `not null` constraint. I don't think we'd want to remove that. I think this might acheive what we want:
`ALTER TABLE water.events ALTER COLUMN event_id DROP DEFAULT;`

src/lib/services/events.js
```diff
@@ -0,0 +1,99 @@
...
+};
+
+/**
+ * Maps data to the format expected by the repository class
+ * @param  {Object} data - event data
+ * @return {Object}      - event data mapped to DB column names/types
+ */
+const mapToRepo = data => {
+  const mapped = mapValues(data, (value, key) => {
+    return isMappedField(value, key) ? JSON.stringify(value) : value;
```
james-siteclick: There is a Bookshelf plugin for defining which fields are JSON:
https://github.com/seegno/bookshelf-json-columns
So JSON.stringify could perhaps be handled at the repository layer using this plugin

src/lib/services/events.js
```diff
@@ -0,0 +1,99 @@
...
+};
+
+/**
+ * Maps keys for data coming from repository class and
+ * returns a new Event model POJO created from the mapped data
+ * @param  {Object} data - data from repository class
+ * @return {Object}      - data with keys camel cased
+ */
+const mapFromRepo = (data) => {
+  const result = keyMapper(data, camelCase);
```
james-siteclick: If using Bookshelf, the keys should hopefully already be camel-cased

src/lib/event.js
```diff
@@ -1,16 +1,9 @@
-const uuidv4 = require('uuid/v4');
 const moment = require('moment');
-const { pool } = require('./connectors/db');
-const Repository = require('@envage/hapi-pg-rest-api/src/repository');
+const repo = require('./connectors/repos');
```
james-siteclick: It feels like this should be a facade to the new service (if possible) rather than connecting to the repo directly

migrations/sqls/20200214162622-event-add-bookshelf-default-timestamp-columns-up.sql
```diff
@@ -0,0 +1,4 @@
+/* Replace with your SQL commands */
+ALTER TABLE water.events ADD COLUMN created_at timestamp DEFAULT now();
+ALTER TABLE water.events ADD COLUMN updated_at timestamp DEFAULT null;
+UPDATE water.events SET created_at = created, updated_at = modified;
```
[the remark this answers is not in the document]
Stephan-Hough: only adding columns for the default Bookshelf managed timestamps. I have deliberately not removed the existing columns because as you say that will break things. Once we have removed the deprecated event service we can remove the modified and updated columns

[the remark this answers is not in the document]
Stephan-Hough: OK i just spotted how you can define custom timestamp columns instead of the default ones. e.g. hasTimestamps: ['date_created', 'date_updated'] Thanks

src/lib/connectors/repos/events.js
```diff
@@ -0,0 +1,45 @@
...
+ */
+const findOne = async (id) => {
+  const model = await Event.forge({ eventId: id }).fetch();
+  return model.toJSON();
+};
+
+/**
+ * Updates the all the fileds for the event record
+ * @param {uuid} id // eventId
+ * @param {Event} event // event model
```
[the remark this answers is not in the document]
Stephan-Hough: done

migrations/sqls/20200214162622-event-add-bookshelf-default-timestamp-columns-up.sql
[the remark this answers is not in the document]
Stephan-Hough: done

src/lib/models/event.js
```diff
@@ -0,0 +1,190 @@
...
+   * Gets the comment for this event
+   * @return {String}
+   */
+  get comment () {
+    return this._comment;
+  }
+
+  /**
+     * Sets the metaData for this event
+     * @param {String} metaData
```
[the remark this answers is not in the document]
Stephan-Hough: done

src/lib/services/events.js
Stephan-Hough: done

src/lib/services/events.js
Stephan-Hough: done

src/lib/event.js
Stephan-Hough: done

src/lib/models/event.js
```diff
@@ -0,0 +1,190 @@
+const Model = require('./model');
```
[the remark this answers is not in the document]
Stephan-Hough: done

PR #786, opened 2020-02-13
Feat: Updates batch creation for better 409 response

swaff:
WATER-2510

In order to provide more detail in the front end, this change updates
the endpoint responsible for creating a new batch so that if there is
already a batch that is processing for the region, a 409 is returned
including the batch that is already processing.

### DIFF
--- a/src/lib/connectors/repos/billing-batches.js
+++ b/src/lib/connectors/repos/billing-batches.js
@@ -13,8 +13,7 @@ const findOne = async (id) => {
     });
 
   return model.toJSON();
-}
-;
+};
 
 const findPage = async (page, pageSize) => {
   const result = await BillingBatch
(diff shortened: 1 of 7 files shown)


src/lib/connectors/repos/billing-batches.js
```diff
@@ -30,6 +29,15 @@ const findPage = async (page, pageSize) => {
...
 
+const findByStatus = async status => {
+  const batches = await BillingBatch
+    .forge()
+    .where({ status })
+    .fetchAll({ withRelated: ['region'] });
+
+  return batches.toJSON();
+};
+
```
james-siteclick: Could we use Bookshelf to find the single pending batch in the region directly, e.g:
```
const findOneByRegionAndStatus = async (regionId, status) => {
  const batch = await BillingBatch
    .forge()
    .where({ region_id : regionId, status })
    .fetch({ withRelated: ['region'], require: false });

  return batch ? batch.toJSON() : null;
};
```

PR #796, opened 2020-02-19
Select returns reminder template by invitation template used

DGajewska:
WATER-2518

* Add call for latest returnInvitation event with all related scheduled notifications which include a specific licence number
* Find appropriate template
* Use control template if no template found


src/lib/connectors/repos/events.js
```diff
@@ -1,12 +1,12 @@
...
+const { Event, bookshelf } = require('../bookshelf');
+const queries = require('./queries/events');
 /**
  * creates a new event record in the database
  * @param {Object} event event pojo
  * @returns {Object} event pojo
  */
 const create = async (event) => {
-  const result = await Event.forge(event).save();
+  const result = await Event.forge().save(event);
```
Stephan-Hough: Interestingly both methods works. I followed the billing batch transactions approach to remain consistent. Did it not work the other way for you?

DGajewska: That's odd. The first way didn't work for the returns notifications feature, it only worked after the change I made

DGajewska: I had a second look at this after I figured out what was going wrong with this event and it does work the way you had it, so I've changed it back

src/modules/batch-notifications/lib/event-helpers.js
```diff
@@ -63,13 +64,18 @@ const updateEventStatus = async (eventId, status, data = {}) => {
 const markAsProcessed = async (eventId, licenceNumbers, recipientCount) => {
   const ev = await evt.load(eventId);
```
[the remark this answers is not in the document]
Stephan-Hough: agreed to do this later

src/modules/batch-notifications/lib/event-helpers.js
```diff
@@ -44,11 +45,11 @@ const createEvent = async (issuer, config, options) => {
  */
 const updateEventStatus = async (eventId, status, data = {}) => {
   const ev = await evt.load(eventId);
```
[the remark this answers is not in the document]
Stephan-Hough: agreed to do this later

PR #797, opened 2020-02-19
Use proxy for cognito requests

james-siteclick:
Fixes an issue where requests to Cognito were not proxied

### DIFF
--- a/config.js
+++ b/config.js
@@ -186,6 +186,7 @@ module.exports = {
 
   cognito: {
     username: process.env.COGNITO_USERNAME,
-    password: process.env.COGNITO_PASSWORD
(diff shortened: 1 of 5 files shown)


james-siteclick:
@swaff I tried your suggestion but I don't think it works unfortunately as the request-promise-native module is a singleton instance, and elsewhere we are doing:
```
require('request-promise-native').defaults({ proxy : null });
```
Therefore I think I need to be explicit about the use of the proxy.
Have had a quick look at other modules.  Some (e.g. axios) allow you to specify a proxy and then no-proxy domains (for communication between services) which would be quite handy for this.  However all the libraries I've looked at so far have various issues.

PR #812, opened 2020-02-26
Enhance CSV upload error handling

DGajewska:
WATER-2534


src/modules/returns/controller.js
```diff
@@ -76,11 +76,11 @@ const patchReturnHeader = async (request, h) => {
...
 /**
  * Creates the event object that represent the upload
- * of a returns xml document.
+ * of a bulk returns document.
  * @param uploadUserName The username of end user.
  * @returns {Object}
  */
-const createXmlUploadEvent = (uploadUserName, subtype = 'xml') => {
+const createBulkUploadEvent = (uploadUserName, subtype = 'csv') => {
   return event.create({
```
Stephan-Hough: This is still referencing the old event service. Do you want to continue with this or use the new event service in lib/services?

Stephan-Hough: agreed to do this later

PR #824, opened 2020-03-04
Add more robust date checking for csv upload.

DGajewska:
WATER-2564

Mapper and validator use same code for dates in csv. Date formats have been made more specific to each return frequency.


src/modules/returns/lib/returns-upload-validator.js
```diff
@@ -101,7 +91,29 @@ const validateReturnlines = ret => {
...
+  if (ret.isNil) {
+    return true;
+  }
+  // if first date line has an invalid date format, frequency won't exist and the
+  // following code will throw an error which surfaces in the UI
+  try {
+    const { startDate, endDate, frequency } = ret;
+    const requiredLines = returnLines.getRequiredLines(startDate, endDate, frequency);
+    const returnTimePeriod = uniq(ret.lines.map(line => line.timePeriod));
+    return returnTimePeriod.length === 1 && returnTimePeriod[0] === requiredLines[0].timePeriod;
```
james-siteclick: It might be that:
```
 return returnTimePeriod === [requiredLines[0].timePeriod];
```
increases the coverage here without needing extra tests

PR #839, opened 2020-03-11
Delete zero transactions

james-siteclick:
- handles 'zero transaction' response from transaction create charge request, and deletes local transaction
- fixes issue where transactions were being generated multiple times
- calls cleanup() for empty and ready batches to remove unwanted licences/invoices


src/lib/connectors/repos/billing-batch-charge-version-years.js
```diff
@@ -1,10 +1,20 @@
...
   BillingBatchChargeVersionYear
     .forge({ billingBatchChargeVersionYearId: id })
     .save(data);
 
+/**
+ * Gets a count of the charge version years in each status by batch ID
+ * @param {String} batchId - guid
+ */
+const findStatusCountsByBatchId = batchId =>
+  raw.multiRow(queries.findStatusCountsByBatchId, { batchId });
```
Stephan-Hough: It's a shame we can't use bookshelf to do this instead of writing SQL

PR #856, opened 2020-03-20
Feat: Persist user errors in bulk upload

DGajewska:
WATER-2584

 - Move final validation step from previewReturn endpoint to pg-boss job after mapping to json.
 - Persist validation results to event metadata. 
 - Delete previewReturn endpoint.


src/modules/returns/lib/jobs/init-upload.js
```diff
@@ -9,12 +10,18 @@ const registerSubscribers = async messageQueue => {
   await messageQueue.subscribe(jobs.mapToJson.jobName, jobs.mapToJson.handler);
 
   await messageQueue.onComplete(jobs.start.jobName, async job => {
-    // XML document has been validated against XSD schema.
-    //
-    // Publish a new job which will convert the valid XML into
-    // a JSON blob which will be uploaded back to S3.
-    const { eventId } = job.data.request.data;
-    await jobs.mapToJson.publish(eventId);
+    if (job.data.failed) return messageQueue.stop();
```
james-siteclick: The `messageQueue` is a singleton instance for the whole application so I don't think it should be stopped.
Possibly all we would wish to do here is set the event to an `error` status and return before further processing.

src/modules/returns/lib/jobs/validate-returns.js
```diff
@@ -0,0 +1,74 @@
...
+*
+* This will acquire the saved document from AWS S3,
+* validate the JSON representation, then put the JSON
+* back to S3 for future processing.
+*
+* @param {Object} job The job data from PG Boss
+*/
+const handleValidateReturnsStart = async job => {
+  const { eventId, companyId } = job.data;
+  const event = await eventsService.findOne(eventId);
```
james-siteclick: If the event is not found, maybe we should throw an error?

src/modules/returns/lib/jobs/validate-returns.js
```diff
@@ -0,0 +1,74 @@
...
+  const jsonData = await loadJson(eventId);
```
james-siteclick: I think this could move inside the try { } block, then if there's a problem the event will be marked in error by the catch { }

src/modules/returns/lib/jobs/init-upload.js
DGajewska: I wasn't sure what I should be doing here, that's why I was keen someone with pg-boss experience had a look at this

PR #857, opened 2020-03-20
Feat/water 2579 annual billing query

james-siteclick:
Modifies the annual billing query to avoid re-processing charge versions in a given financial year.
Moves code for creation of billing_batch_charge_versions to a new repo and fixes camel-casing.


src/lib/connectors/bookshelf/ChargeVersion.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = bookshelf.model('ChargeVersion', {
+  tableName: 'charge_versions',
+
+  hasTimestamps: ['date_created', 'date_updated'],
+
+  idAttribute: 'charge_version_id',
+
+  chargeElements () {
+    return this.hasMany('ChargeElement', 'charge_version_id', 'charge_version_id');
```
Stephan-Hough: 'charge_version_id' has been added twice.

src/lib/connectors/repos/queries/billing-batch-charge-versions.js
```diff
@@ -0,0 +1,70 @@
...
+  and l.region_id = :regionId::uuid
+  and (cv.end_date is null or cv.end_date > :fromDate)
+returning *;`;
+
+exports.createAnnual = `
+    insert into water.billing_batch_charge_versions (billing_batch_id, charge_version_id)
+    select :billingBatchId, cv.charge_version_id
+    from water.licences l
+      join water.charge_versions cv on l.licence_ref = cv.licence_ref
+      join water.charge_elements ce on cv.charge_version_id = ce.charge_version_id
```
Stephan-Hough: Do you need to join the charge element table here? Is it used as a constraint?

src/lib/connectors/repos/queries/billing-batch-charge-versions.js
```diff
@@ -0,0 +1,70 @@
...
+      )
+    returning *;`;
+
+exports.createTwoPartTariff = `
+      insert into water.billing_batch_charge_versions (billing_batch_id, charge_version_id)
+      select :billingBatchId, cv.charge_version_id
+      from water.licences l
+        join water.charge_versions cv on l.licence_ref = cv.licence_ref
+        join water.licence_agreements la on l.licence_ref = la.licence_ref
+        join water.charge_elements ce on cv.charge_version_id = ce.charge_version_id
```
Stephan-Hough: Same as above comment.

src/lib/connectors/bookshelf/ChargeVersion.js
james-siteclick: Once is for the local table and once for the foreign key - I think this is OK

src/lib/connectors/repos/queries/billing-batch-charge-versions.js
james-siteclick: It ensures any charge versions won't be processed if they don't contain any charge elements, so is quite handy from that respect

src/lib/connectors/repos/queries/billing-batch-charge-versions.js
james-siteclick: I think in this case it ensures we only process charge versions which have a charge element matching the season of the batch type (summer or winter/all year)

src/lib/connectors/repos/queries/billing-batch-charge-versions.js
james-siteclick: I think here it ensures we only process charge versions which have charge elements

PR #873, opened 2020-03-30
API for getting one invoice licence with transactions

Stephan-Hough:
feat/water-2479
--route, controller, service and repo methods created
-- invoiceLicence bookshelf id attribute added


src/modules/billing/controller.js
```diff
@@ -161,13 +161,23 @@ const getBatchLicences = async (request, h) => {
   return invoiceLicenceService.getLicencesWithTransactionStatusesForBatch(batch.id);
 };
 
+const getInvoiceLicenceWithTransactions = async (request, h) => {
+  const { invoiceLicenceId } = request.params;
+  try {
+    const invoiceLicence = await invoiceLicenceService.getInvoiceLicenceWithTransactions(invoiceLicenceId);
+    return invoiceLicence;
+  } catch (err) {
+    return err;
```
james-siteclick, line 170: We seem to be swallowing this error here so we may as well remove the try/catch.
Also maybe not handling case where invoiceLicence is not found.  Perhaps this should just be:
```
const { invoiceLicenceId } = request.params;
const invoiceLicence = await invoiceLicenceService.getInvoiceLicenceWithTransactions(invoiceLicenceId);
return invoiceLicence || Boom.notFound(`Invoice licence ${ invoiceLicenceId } not found`);
```

May need to use the `require: false` option in the Bookshelf repo for this to work, as by default Bookshelf will throw an error if the model is not found.

Stephan-Hough: Done

PR #877, opened 2020-04-02
Incorporate returns matching

DGajewska:
WATER-2421

- Add two part tariff batch handling in process-charge-version job
- For each invoice licence in the batch:
  - Get returns for each licence
  - Map charge elements from each transaction
  - Pass returns and charge elements into returns matching algorithm
- Map matching results to Transaction
- Update the batch status to `review` when all licences have been processed


src/modules/billing/services/two-part-tariff-service/index.js
```diff
@@ -0,0 +1,91 @@
+const { set } = require('lodash');
+const twoPartTariffMatching = require('./two-part-tariff-matching');
+const returnHelpers = require('./returns-helpers');
+const Purpose = require('../../../../lib/models/purpose');
+
+const mapResultsWithData = (overallError, data, transactions) => {
+  const updatedTransactions = [];
+  for (const result of data) {
+    const { error, data: { chargeElementId, actualReturnQuantity } } = result;
+    const [transaction] = transactions.filter(transaction => transaction.chargeElement.id === chargeElementId);
```
james-siteclick, line 10: If you only need the first element, may be neater to use arr.find() instead of arr.filter(), e.g.:
```
const transaction = transactions.find(...
```

src/modules/billing/services/two-part-tariff-service/index.js
```diff
@@ -0,0 +1,91 @@
...
+
+    const twoPartTariffStatus = overallError || error;
+    set(transaction, 'calculatedVolume', actualReturnQuantity);
```
james-siteclick, line 13: The service models have a `fromHash` method for setting lots of properties, e.g.:
```
transaction.fromHash({
   volume : twoPartTariffStatus ? null : actualReturnQuantity,
   twoPartTariffStatus,
  ...
```

src/modules/billing/services/two-part-tariff-service/index.js
```diff
@@ -0,0 +1,91 @@
+const { set } = require('lodash');
+const twoPartTariffMatching = require('./two-part-tariff-matching');
+const returnHelpers = require('./returns-helpers');
+const Purpose = require('../../../../lib/models/purpose');
+
+const mapResultsWithData = (overallError, data, transactions) => {
```
james-siteclick, line 6: Because the transactions passed in are not cloned, I think these functions are mutating the transactions passed in. 
In this case it should perhaps be named `decorateTransactions` or similar.

src/modules/billing/services/two-part-tariff-service/index.js
```diff
@@ -0,0 +1,91 @@
...
+  const updatedTransactions = [];
```
james-siteclick, line 7: I don't think this array is doing anything, I think if the function returned `transactions` at the end it may have the same effect?|

src/modules/billing/controller.js
```diff
@@ -30,16 +30,24 @@ const createBatchEvent = (userEmail, batch) => {
   return eventService.create(batchEvent);
 };
 
+const getFromFinancialYearEndingForBatch = (batchType, financialYearEnding) => {
```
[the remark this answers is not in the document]
DGajewska: Yes, I've already moved this on Friday as I noticed it when I was reviewing your PR :)

src/modules/billing/services/two-part-tariff-service/index.js
DGajewska: You're right. I thought I was receiving a copy of the transaction from filter, but I guess not. I should've double checked this, but I will know for next time!

src/modules/billing/controller.js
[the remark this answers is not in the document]
DGajewska: Yes, I realised that it would conflict when I was reviewing your PR on Friday

PR #886, opened 2020-04-07
Add twoPartTariffError flag to batch licences API

james-siteclick:
Since the twoPartTariffStatuses will not be cleared when a volume is set by the user, we need the twoPartTariffError flag adding to the batch licence list API.

Also fixes an issue mapping transactions from DB rows where the twoPartTariffError flag is null

### DIFF
--- a/src/modules/billing/mappers/transaction.js
+++ b/src/modules/billing/mappers/transaction.js
@@ -2,7 +2,7 @@
 
 const moment = require('moment');
 const { titleCase } = require('title-case');
-const { get, pick, identity } = require('lodash');
+const { get, pick, identity, pickBy, isBoolean } = require('lodash');
 
 const Batch = require('../../../lib/models/batch');
 const DateRange = require('../../../lib/models/date-range');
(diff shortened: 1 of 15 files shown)


src/modules/billing/mappers/transaction.js
```diff
@@ -148,13 +148,16 @@ const dbToModel = row => {
...
     chargePeriod: new DateRange(row.startDate, row.endDate),
     isCompensationCharge: row.chargeType === 'compensation',
     chargeElement: chargeElementMapper.dbToModel(row.chargeElement),
     volume: parseFloat(row.volume),
     agreements: mapDBToAgreements(row),
-    twoPartTariffReview: mapReviewDataToUser(row.twoPartTariffReview)
+    twoPartTariffReview: mapReviewDataToUser(row.twoPartTariffReview),
+    ...pickBy({
+      twoPartTariffError: row.twoPartTariffError
+    }, isBoolean)
```
Stephan-Hough, line 160: Note we added an assertIsNullableBoolean to the transaction model for twoPartTariffError so no need to make the change here however this is a belt and braces approach so could probably stay as is.

PR #889, opened 2020-04-08
Creates API for updating transaction volume

DGajewska:
WATER-2544

Updates transaction volume during two part tariff review.

This includes updating the two part tariff error to false and storing the user who updated the volume.


src/modules/billing/controller.js
```diff
@@ -161,6 +163,23 @@ const getBatchLicences = async (request, h) => {
...
+  const { internalCallingUser: user } = request.defra;
+
+  const batch = await transactionsService.getById(transactionId);
+  if (!batch) return Boom.notFound(`No transaction (${transactionId}) found`);
+
+  checkVolumeUpdateCriteriaMet(batch, volume);
+
+  const updatedTransaction = decorateTransactionWithVolume(batch, volume, user);
+
+  await transactionsService.updateTransactionVolume(updatedTransaction);
```
Stephan-Hough, line 178: What if the update fails, should we use boom here as well?

src/modules/billing/lib/transaction.js
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * Throws Boom.badRequest if the validator returns an error
+ *
+ * @param  {Batch} batch   containing the transaction
+ * @param  {Integer} volume   to update transaction with
+ */
+const checkVolumeUpdateCriteriaMet = (batch, volume) => {
+  const { invoices: [{ invoiceLicences: [{ transactions: [transaction] }] }] } = batch;
+  const error = validator(batch, transaction, volume);
+  if (error) throw Boom.badRequest(error);
```
james-siteclick, line 64: I wonder if the Boom errors should only really be in the controller

src/lib/models/transaction.js
```diff
@@ -196,6 +196,12 @@ class Transaction extends Model {
 
   set volume (volume) {
     validators.assertNullableQuantity(volume);
```
james-siteclick, line 198: All the validation can be rolled into a single validator here:
```
// validators.js
const assertNullableQuantityWithMaximum(value, max) =>
  assert(value, VALID_NULLABLE_QUANTITY.max(max));
```
And we can make use of the chargeElement.volume property which already does billableAnnualQuantity || authorisedAnnualQuantity:
```
// transaction.js
assertNullableQuantityWithMaximum(value, this.chargeElement.volume);
this._volume = volume;
```

src/modules/billing/services/transactions-service.js
```diff
@@ -78,7 +79,46 @@ const setErrorStatus = transactionId =>
...
+ *
+ * Throws error with corresponding message if criteria is not met
+ *
+ * @param  {Batch} batch   for the transaction
+ * @param  {Transaction} transaction   in question
+ * @param  {Integer} volume   to update transaction with
+ * @param  {Object} user   id and email of internal user making the update
+ */
+const updateTransactionVolume = async (batch, transaction, volume, user) => {
+  const flags = [batchIsTwoPartTariff(batch), batchIsInReviewStatus(batch), transactionIsCandidate(transaction)];
```
james-siteclick, line 107: I think since this work was reviewed originally, we've introduced some custom errors which it would be nice to throw here.
There is a BatchStatusError in src/modules/billing/lib/errors.js.
We can introduce a TransactionStatusError for the third one in src/modules/billing/lib/errors.js which also extends StateError.
 The benefit of these is that it's easier in consuming code to do, e.g `err instanceof BatchStatusError`, and it's also possible to check if they extend a certain base class, e.g. `err instanceof StateError` would also return true.

src/lib/models/transaction.js
DGajewska: I did think to use a validator here, I can't remember why I decided against it

DGajewska: If the transaction had a matching error, the volume will have been set to `null` at this point, so I have had to find the max from the two quantities

PR #898, opened 2020-04-16
Adds endpoint for updating tpt bill run to ready status.

DGajewska:
WATER-2600

Checks that batch is in review status.
Checks that all twoPartTariffErrors have been resolved.
Updates batch status to processing.
Saves an 'approve-review' event.
Kicks off prepare-transactions job to continue bill run process.

### DIFF
--- a/src/modules/billing/services/batch-service.js
+++ b/src/modules/billing/services/batch-service.js
@@ -274,11 +274,23 @@ const getTransactionsWithTwoPartError = batch => {
     invoice.invoiceLicences.forEach(invoiceLicence => {
       invoiceLicence.transactions.forEach(transaction => {
         if (transaction.twoPartTariffError) acc.push(transaction);
-        return acc;
       });
     });
     return acc;
   }, []);
+};
+
+const assertNoTransactionsWithTwoPartError = batch => {
+  const transactions = getTransactionsWithTwoPartError(batch);
+  if (transactions.length) {
+    throw new TransactionStatusError('Cannot approve review. There are outstanding two part tariff errors to resolve');
+  }
+};
+
+const assertBatchStatusIsReview = batch => {
+  if (batch.status !== BATCH_STATUS.review) {
+    throw new BatchStatusError('Cannot approve review. Batch status must be "review"');
+  }
 };
 
 /**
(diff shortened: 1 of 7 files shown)


src/modules/billing/services/batch-service.js
```diff
@@ -268,6 +269,47 @@ const createChargeModuleBillRun = async batchId => {
...
+    invoice.invoiceLicences.forEach(invoiceLicence => {
+      invoiceLicence.transactions.forEach(transaction => {
+        if (transaction.twoPartTariffError) acc.push(transaction);
+        return acc;
+      });
+    });
+    return acc;
+  }, []);
+};
+
```
james-siteclick, line 283: I think I'd find this a bit easier to understand if we split the logic out into some extra functions:
```
const assertNoTransactionsWithTwoPartError = batch => {
  const transactions = getTransactionsWithTwoPartError(batch);
  if(transactions.length) {
    throw new TransactionStatusError('Cannot approve review. There are outstanding two part tariff errors to resolve');
   }
}

const assertBatchStatusIsReview = batch => {
  if(batch.status !== BATCH_STATUS.review) {
    throw new BatchStatusError('Cannot approve review. Batch status must be "review"');
  } 
}
```

Then the function below can do:
```
const approveTptBatchReview = async batch => {
  assertNoTransactionsWithTwoPartError(batch);
  assertBatchStatusIsReview(batch);
  await setStatus(batch.id, BATCH_STATUS.processing);
  return getBatchById(batch.id);
}
```

src/modules/billing/services/batch-service.js
```diff
@@ -268,6 +269,47 @@ const createChargeModuleBillRun = async batchId => {
   return batch.pickFrom(row, ['externalId', 'billRunId']);
 };
 
+const getTransactionsWithTwoPartError = batch => {
+  return batch.invoices.reduce((acc, invoice) => {
+    invoice.invoiceLicences.forEach(invoiceLicence => {
+      invoiceLicence.transactions.forEach(transaction => {
+        if (transaction.twoPartTariffError) acc.push(transaction);
+        return acc;
```
james-siteclick, line 277: I'm not certain this return is necessary

PR #903, opened 2020-04-21
Fixes snags in TPT flow following review.

DGajewska:
WATER-2598

Fixes mapping issue with transaction after volume update.
Updates batch event when review stage is reached.
Refactors User mappers to own file.


src/modules/billing/services/job-service.js
```diff
@@ -37,6 +37,15 @@ const setEmptyBatch = (eventId, batchId) =>
 const setFailedJob = (eventId, batchId) =>
   setStatuses(eventId, jobStatus.error, batchId, BATCH_STATUS.error);
 
+/**
+* Sets the event/job to error, and the batch to error
+* @param {String} eventId
+* @param {String} batchId
+*/
+const setReviewJob = (eventId, batchId) =>
+  setStatuses(eventId, jobStatus.review, batchId, BATCH_STATUS.review);
```
james-siteclick, line 46: I think we need to address how the events are used here possibly in a future ticket/refinement

DGajewska: I've made a list of things of possible bugs/things that might need changing, I'll add it to that list. The waiting page relies on the event status and changing that felt too big and too out of scope of this ticket.

PR #932, opened 2020-05-14
Feat: Decouple calculatedVolumes from Transaction

DGajewska:
WATER-2658

Migrations to create billing_volumes table and drop tpt colums in billing_transactions.
Create BillingVolume repo.
Create BillingVolume model.
Create billingVolume service.
Create billingVolume mappers.
Remove tpt references in Transaction model and mappers.
Add placeholder function in two part tariff service interface for new function.


src/modules/billing/services/billing-volumes-service.js
```diff
@@ -0,0 +1,82 @@
...
+  return chargeElementIsSummer === isSummer;
+};
+
+const getChargeElementIds = (chargeVersion, isSummer) => {
+  return chargeVersion.chargeElements.filter(chargeElement => {
+    if (chargeElementSeasonMatches(chargeElement, isSummer)) {
+      return chargeElement.chargeElementId;
+    };
+  });
+};
```
james-siteclick, line 17: I struggled a bit to understand this, and I don't think it is returning an array of IDs.
Maybe this might be simpler:
```
const isSummerChargeElement = chargeElement => chargeElement.season === 'summer';

const getChargeElementsForSeason = (chargeVersion, isSummer) => {
  return chargeVersion.chargeElements
   .filter(chargeElement => isSummerChargeElement(chargeElement) === isSummer);
}
```

src/modules/billing/services/billing-volumes-service.js
```diff
@@ -0,0 +1,82 @@
...
+const volumesToModels = volumes => volumes.map(billingVolumeToModel);
+
+const isMissingVolumes = volumes => {
+  const containsNull = volumes.map(volume => isEmpty(volume));
+  return containsNull.includes(true);
+};
+
+const getVolumesForChargeElements = async (chargeVersion, financialYear, isSummer) => {
+  const chargeElements = getChargeElementIds(chargeVersion, isSummer);
+  const billingVolumes = await Promise.all(
```
james-siteclick, line 30: We maybe want a repo method here that can accept an array of IDs (this can be done with Bookshelf collections if fetching related models) to get all the data in one go.
```
const chargeElementIds = chargeElements.map(chargeElement => chargeElement.chargeElementId);
const billingVolumes = await billingVolumesRepo.findByChargeElementIdsFinancialYearAndSeason(chargeElementIds, financialYear, isSummer);
```

src/modules/billing/services/billing-volumes-service.js
```diff
@@ -0,0 +1,82 @@
...
+    chargeElements.map(async chargeElement => {
+      return billingVolumesRepo.findByChargeElementId(chargeElement.chargeElementId);
+    }));
+  if (isMissingVolumes(billingVolumes)) return null;
```
james-siteclick, line 34: I think if this is checking there are no missing volumes we can then do
```
if(chargeElements.length !== billingVolumes.length)
```

migrations/sqls/20200507130236-billing-volumes-table-up.sql
```diff
@@ -0,0 +1,19 @@
...
+    constraint billing_volumes_charge_element_id_fkey
+    references water.charge_elements (charge_element_id),
+  financial_year smallint not null,
+  is_summer boolean not null,
+  calculated_volume numeric,
+  two_part_tariff_error boolean not null default false,
+  two_part_tariff_status integer,
+  two_part_tariff_review jsonb,
+  is_approved boolean not null default false
+);
```
james-siteclick, line 13: I think this table may need a unique constraint on (charge_element_id, financial_year, is_summer)

src/modules/billing/services/billing-volumes-service.js
```diff
@@ -0,0 +1,82 @@
...
+
+  const relevantVolumes = flatMap(billingVolumes).filter(volume => volume.financialYear === financialYear && volume.isSummer === isSummer);
```
james-siteclick, line 36: I think with the repo change above this step may not be necessary

src/modules/billing/services/billing-volumes-service.js
DGajewska: This was a map method, but I clearly changed it and didn't update the return value

src/lib/connectors/repos/billing-volumes.js
```diff
@@ -0,0 +1,40 @@
...
+  return result.toJSON();
+};
+
+/**
+ * Updates the record with the fields contained within changes object
+ * @param {String} id
+ * @param {Object} changes
+ */
+const update = async (id, changes) => {
+  const result = await BillingVolume.forge({ billingEventId: id }).save(changes);
```
james-siteclick, line 34: Should this be:
```
.forge({ billingVolumeId: id })
```

src/modules/billing/services/billing-volumes-service.js
```diff
@@ -0,0 +1,65 @@
+const billingVolumesRepo = require('../../../lib/connectors/repos/billing-volumes');
+const mappers = require('../mappers');
+const twoPartTariffMatching = require('./two-part-tariff-service');
+
+const isSummerChargeElement = chargeElement => chargeElement.season === 'summer';
+
+const getChargeElementsForSeason = (chargeVersion, isSummer) =>
+  chargeVersion.chargeElements.filter(chargeElement => isSummerChargeElement(chargeElement) === isSummer);
+
+const billingVolumeToModel = volume => mappers.billingVolume.dbToModel(volume);
```
james-siteclick, line 10: This extra function doesn't seem to be doing anything, e.g.
```
const volumesToModels = volumes => volumes.map(mappers.billingVolume.dbToModel);
```

src/modules/billing/services/billing-volumes-service.js
```diff
@@ -0,0 +1,65 @@
...
+
+const volumesToModels = volumes => volumes.map(billingVolumeToModel);
+
+const getVolumesForChargeElements = async (chargeVersion, financialYear, isSummer) => {
+  const chargeElements = getChargeElementsForSeason(chargeVersion, isSummer);
+  const chargeElementIds = chargeElements.map(chargeElement => chargeElement.chargeElementId);
+  const billingVolumes = await billingVolumesRepo.findByChargeElementIdsAndFinancialYear(chargeElementIds, financialYear);
+
+  if (chargeElements.length !== billingVolumes.length) return null;
```
james-siteclick, line 19: Do you think there are 3 possible cases here we need to handle:
```
// The matching process has never run - no billing volumes are persisted
if(billingVolumes.length === 0) {
  return null;
}
// There is an exact match, it has run OK
if(chargeElements.length === billingVolumes.length) {
  return volumesToModels(billingVolumes);
}
// There is a mismatch - something has gone wrong
throw new NotFoundError(`Billing volumes missing for charge version ${ chargeVersion.id }`);
```

PR #971, opened 2020-06-05
API to rename licence and create event for KPI data

Stephan-Hough:
feat/water-2644
-- postLicenceName route added to rename licence
-- controller handler post method added
-- crm documents method to update documentHeader
-- tests added and updated


src/modules/licences/controller.js
```diff
@@ -292,10 +292,27 @@ const getLicenceCompanyByDocumentId = async (request, h) => {
   }
 };
 
+const postLicenceName = async (request, h) => {
+  const { documentId } = request.params;
+  const { documentName, rename } = request.payload;
```
james-siteclick, line 297: I think the value of `rename` can be established without needing an extra payload parameter, along the lines of:
```
const { isEmpty } = require('lodash');
// ...
const { data : { document_name : currentName } } = await documentsClient.findOne(documentId);
const rename = !isEmpty(currentName);
```

src/modules/licences/routes.js
```diff
@@ -126,5 +126,24 @@ module.exports = {
...
+    handler: controller.postLicenceName,
+    config: {
+      description: 'Sets name for a licence',
+      validate: {
+        params: {
+          documentId: Joi.string().guid().required()
+        },
+        payload: {
+          documentName: Joi.string().required().allow(''),
+          rename: Joi.boolean().required(),
```
james-siteclick, line 143: As above, not sure the rename param is needed

src/modules/licences/controller.js
Stephan-Hough: As discussed in the SLACK the logic should not sit with the UI so this has been implemented as requested

src/modules/licences/routes.js
Stephan-Hough: done

PR #977, opened 2020-06-08
Feature/2682 import licence versions

swaff:


migrations/sqls/20200605100709-licence-versions-unique-constraint-up.sql
```diff
@@ -0,0 +1,14 @@
+alter table water.licence_versions
+  add constraint uindx_licence_versions_external_id unique (external_id);
+
+alter table water.licence_version_purposes
```
james-siteclick, line 4: If I've understood correctly, I don't think this unique constraint is valid, e.g. there may be many records for a licence version with the same purpose and abs period.
I tried to verify this with:
```
select p."AABV_AABL_ID", p."AABV_ISSUE_NO", p."AABV_INCR_NO", p."FGAC_REGION_CODE" , 
concat_ws('-', p."APUR_APPR_CODE", p."APUR_APSE_CODE", p."APUR_APUS_CODE") as purposes,
concat_ws('-', p."PERIOD_ST_DAY", p."PERIOD_ST_MONTH", p."PERIOD_END_DAY", p."PERIOD_END_MONTH") as abs_period,
COUNT(*) as purpose_count
from import."NALD_ABS_LIC_PURPOSES" p
group by p."AABV_AABL_ID", p."AABV_ISSUE_NO", p."AABV_INCR_NO", p."FGAC_REGION_CODE", 
concat_ws('-', p."APUR_APPR_CODE", p."APUR_APSE_CODE", p."APUR_APUS_CODE"), 
concat_ws('-', p."PERIOD_ST_DAY", p."PERIOD_ST_MONTH", p."PERIOD_END_DAY", p."PERIOD_END_MONTH")
order by count(*) desc
```

PR #979, opened 2020-06-09
Feat: Update TPT Process

DGajewska:
WATER-2664

* Adds batchId to billing_volumes table
* Calls billingVolumes service from chargeProcessor
* Updates twoPartTariff service to work with new TPT process
* Checks billing_volumes table for unapproved billingVolumes
* Sets batch status to review if there are unapproved billingVolumes
* Approves billingVolumes when a review batch is approved

### DIFF
--- a/src/modules/billing/services/two-part-tariff-service/prepare-returns.js
+++ b/src/modules/billing/services/two-part-tariff-service/prepare-returns.js
@@ -80,11 +80,8 @@ const checkDueReturnsErrors = returns => {
   }
 };
 
-const checkReturnsAndLines = returns => {
-  if (returns.length === 0) return true;
-  const returnsWithLines = returns.map(ret => !!ret.lines);
-  return returnsWithLines.includes(false);
-};
+const checkReturnsExistAndHaveLines = returns =>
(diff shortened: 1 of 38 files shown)


src/modules/billing/services/two-part-tariff-service/prepare-returns.js
```diff
@@ -80,6 +80,12 @@ const checkDueReturnsErrors = returns => {
   }
 };
 
+const checkReturnsAndLines = returns => {
+  if (returns.length === 0) return true;
+  const returnsWithLines = returns.map(ret => !!ret.lines);
+  return returnsWithLines.includes(false);
+};
+
```
swaff, line 88: I think this function name could be improved to indicate what is being checked for

It might also be able to be shortened to 

```
return returns.length === 0
  ? true
  : returns.some(ret => !ret.lines)
```

DGajewska: Thank Mark :) I have made the suggested changes and have added a test for this scenario as I realised that I didn't add a test for a the missing lines scenario

PR #995, opened 2020-06-19
Feat: Backend Updates for billingVolumes in UI

DGajewska:
WATER-2699

* Adds volume column to billing_volumes table
* Adds volume to billingVolumes model and mappers
* Updates patchTransaction to only update relevant billingVolume
* Updates TPT approval to use billingVolumes. Adds updating transactin to prepare-transactions job.
* Adds billingVolumes to queries, mappers, and models where required in the UI

### DIFF
--- a/src/lib/connectors/repos/billing-batches.js
+++ b/src/lib/connectors/repos/billing-batches.js
@@ -6,6 +6,18 @@ const { paginatedEnvelope } = require('./lib/envelope');
 
 const mapModel = model => model ? model.toJSON() : null;
 
+const findRelevantBillingVolumes = billingBatch => {
+  for (const invoice of billingBatch.billingInvoices) {
+    for (const invoiceLicence of invoice.billingInvoiceLicences) {
+      for (const transaction of invoiceLicence.billingTransactions) {
+        transaction.chargeElement.billingVolume = transaction.chargeElement.billingVolume.find(
+          billingVolume => billingVolume.billingBatchId === billingBatch.billingBatchId);
+      }
+    }
+  }
+  return billingBatch;
+};
+
 const findOne = async (id) => {
   const model = await BillingBatch
     .forge({ billingBatchId: id })
@@ -79,7 +91,7 @@ const findOneWithInvoices = async (id) => {
 };
 
 const findOneWithInvoicesWithTransactions = async (id) => {
-  const model = await BillingBatch
+  const results = await BillingBatch
     .forge({ billingBatchId: id })
     .fetch({
       withRelated: [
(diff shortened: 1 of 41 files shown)


src/modules/billing/mappers/transaction.js
```diff
@@ -56,6 +57,9 @@ const dbToModel = row => {
     volume: row.volume ? parseFloat(row.volume) : null,
     agreements: mapDBToAgreements(row)
   });
+  if (row.billingVolume) {
```
james-siteclick, line 60: If row.billingVolumes had been left as an array, would it be possible find the correct one by looking at the transaction season and the financial year of its charge period? 

I think the repo should probably return the data with relationships that match the DB schema, and the mapping to a single billingVolume should take place here

src/lib/connectors/repos/billing-batches.js
```diff
@@ -89,12 +101,17 @@ const findOneWithInvoicesWithTransactions = async (id) => {
         'billingInvoices.billingInvoiceLicences.licence',
         'billingInvoices.billingInvoiceLicences.licence.region',
         'billingInvoices.billingInvoiceLicences.billingTransactions',
+        'billingInvoices.billingInvoiceLicences.billingTransactions.billingVolume',
```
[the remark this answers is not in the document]
james-siteclick: Upon investigation these don't seem possible with Bookshelf, so we are relying on the mapping. layer for now

src/modules/billing/mappers/transaction.js
```diff
@@ -40,23 +41,38 @@ const mapDBToAgreements = row => {
   return agreements.filter(identity);
 };
 
+const doesVolumeMatchTransaction = (volume, transaction) => {
+  const transactionFinacialYear = transaction.endDate.slice(0, 4);
```
james-siteclick, line 45: I think this needs to consider the case where the charge period is not the full financial year. 
I think there is a helper for this:
```
const helpers = require('@envage/water-abstraction-helpers');
const transactionFinancialYear = helpers.charging.getFinancialYear(transaction.endDate);
```

src/modules/billing/mappers/transaction.js
```diff
@@ -40,23 +41,38 @@ const mapDBToAgreements = row => {
...
+  const isSummerTransaction = transaction.season === 'summer';
+
+  const financialYearsMatch = parseInt(volume.financialYear) === parseInt(transactionFinacialYear);
```
james-siteclick, line 48: There is a little typo:
transactionFin**ac**ialYear

PR #1004, opened 2020-06-25
Feat: Charge data for licence versions

swaff:
Dependency: https://github.com/DEFRA/water-abstraction-import/pull/182

WATER-2731

 - Adds the loss factor to the water.purposes_uses table
 - Adds a new route for the charge elements endpoint
 - Returns 404 if the licence version is not found
 - Updates water.purposes_uses with is_two_part_tariff flag.
 - Updates the PurposeUse model to include the isTwoPartTariff flag.
 - Updates the transaction process to use the new flag rather than having to calculate the vale from the purpose codes.
 - Updates the abstraction period model to compare itself to other abstraction periods
 - Adds the description to the ChargeElement model
 - Updates the charge element controller to return the mapper ChargeElements for the licence version


src/lib/models/abstraction-period.js
```diff
@@ -1,7 +1,55 @@
+'use strict';
+
+const { CHARGE_SEASON } = require('./constants');
 const Model = require('./model');
 const { assertDay, assertMonth } = require('./validators');
 
+const moment = require('moment');
```
james-siteclick, line 7: There could be shorter solution to this if we used moment range which we've used elsewhere, e.g.:
```
const moment = require('moment-range').extendMoment(require('moment'));

/**
 * Creates a moment range from the supplied abs period in the specified year
 * @param {AbstractionPeriod} abstractionPeriod
 * @param {Number} startYear
 * @return {MomentRange}
 */
const createRange = (abstractionPeriod, startYear) => {
  const { startDay, endDay, startMonth, endMonth } = abstractionPeriod;
  const isCrossYear = (startMonth > endMonth || (startMonth === endMonth && startDay > endDay));
  const endYear = isCrossYear ? startYear + 1 : startYear;
  return moment.range(
    `${startYear}-${startMonth}-${startDay}`,
    `${endYear}-${endMonth}-${endDay}`
  );
};

...

  /**
   * Checks if the passed AbstractionPeriod contains this instance. In order
   * for this to return true, this instance must fit inside the passed
   * AbstractionPeriod, but not matching the bounaries.
   *
   * @param {AbstractionPeriod} period The abstraction period to check if this instance fits within
   */
  isWithinAbstractionPeriod (period) {
    const thisRange = createRange(this, 2018);
    const testRanges = [createRange(period, 2017), createRange(period, 2018)];
    return testRanges.some(range => range.contains(thisRange));
  }
```

src/lib/models/abstraction-period.js
```diff
@@ -53,6 +101,80 @@ class AbstractionPeriod extends Model {
...
+    this.startDay = startDay;
+    this.startMonth = startMonth;
+    this.endDay = endDay;
+    this.endMonth = endMonth;
+  }
+
+  /**
+   * Checks if the passed AbstractionPeriod contains this instance. In order
+   * for this to return true, this instance must fit inside the passed
+   * AbstractionPeriod, but not matching the bounaries.
```
james-siteclick, line 123: I'm fairly sure matching the boundaries should also be included.
E.g. abs period from 1/4 to 31/10 would count as summer etc.

src/lib/models/date-range.js
```diff
@@ -3,6 +3,11 @@
 const validators = require('./validators');
 
 class DateRange {
+  /**
+   *
+   * @param {String} startDate The start date in the format YYYY-MM-DD
+   * @param {String?} endDate Optional end date in the format YYYY-MM-DD
```
james-siteclick, line 9: If we're using JSDoc syntax optional params are shown as:
```
  * @param {String} [endDate] Optional end date in the format YYYY-MM-DD
```

PR #1027, opened 2020-07-07
Feat: Create invoice account with address, agent & contact roles.

DGajewska:
WATER-2716

Creates routes and controller functions.
Validates all entities before any futher processing is completed.
Calls CRM to create new address, agent & contact entities as required.


src/modules/billing/mappers/contact.js
```diff
@@ -3,6 +3,14 @@
 const { isEmpty } = require('lodash');
 const Contact = require('../../../lib/models/contact-v2');
 
+const getInitials = contact => {
```
james-siteclick, line 6: I think it would give more flexibility to add a `middleInitials` property to the Contact model above rather than map everything to the existing `initials` property

src/lib/models/contact-v2.js
```diff
@@ -40,13 +42,41 @@ class Contact extends Model {
...
+
   /**
-   * Gets the contact's full name including salutation, first name/initial and surname
+   * Gets the contact's full name including title, first name/initial, surname and suffix
    * @return {String}
    */
   get fullName () {
-    const parts = [this._salutation, this._initials || this._firstName, this._lastName];
-    return parts.filter(x => x).join(' ');
+    const parts = [this._title, this._initials || this._firstName, this._lastName];
```
james-siteclick, line 77: This would need to handle initials/middle initials.

I think we would also need to store the `dataSource` property so we could decide how to map the name.

src/modules/companies/controller.js
```diff
@@ -44,4 +50,39 @@ const getReturns = async (request, h) => {
...
+};
+
+/**
+ * Gets all addresses for a company
+ */
+const getCompanyAddresses = async (request, h) => {
+  const { companyId } = request.params;
+  const companyAddresses = await crmConnectors.companies.getCompanyAddresses(companyId);
+  if (!isEmpty(companyAddresses)) return companyAddresses;
+  throw Boom.notFound(`Addresses for company ${companyId} not found`);
```
james-siteclick, line 70: If addresses is an empty array I would expect this to be returned with a `200`

I'd only expect `404` in the case where the company is not found

src/modules/companies/controller.js
```diff
@@ -44,4 +50,39 @@ const getReturns = async (request, h) => {
...
 };
 
+/**
+ * Gets a company for the specified ID
+ */
+const getCompany = async (request, h) => {
+  const { companyId } = request.params;
+  const company = await crmConnectors.companies.getCompany(companyId);
+  if (company) return company;
+  throw Boom.notFound(`Company ${companyId} not found`);
```
james-siteclick, line 60: I think we've been generally returning Boom errors rather than throwing (at least recently...)

src/modules/companies/controller.js
```diff
@@ -44,4 +50,39 @@ const getReturns = async (request, h) => {
   return returns.map(rowMapper);
 };
 
+/**
+ * Gets a company for the specified ID
+ */
+const getCompany = async (request, h) => {
+  const { companyId } = request.params;
+  const company = await crmConnectors.companies.getCompany(companyId);
```
james-siteclick, line 58: Are the controller methods here returning data in the service model shape, or directly from the CRM?  (It should be the service models)

src/modules/companies/lib/invoice-accounts.js
```diff
@@ -0,0 +1,146 @@
+const waterServiceRepos = require('../../../lib/connectors/repos');
```
james-siteclick, line 1: I think the code in here should probably be merged with `src/modules/billing/services/invoice-account.js` and placed in the global location:

`src/lib/services`

Although if creating other entities (e.g. company, address) these should belong in the services specific to those entities.

src/modules/companies/lib/invoice-accounts.js
```diff
@@ -0,0 +1,146 @@
...
+const crmConnectors = require('../../../lib/connectors/crm-v2');
+
+const invoiceAccountsValidators = require('../validators/invoice-accounts');
+const moment = require('moment');
+const { omit, startCase } = require('lodash');
+
+const getAgentCompany = async agent => {
+  if (agent.companyId) return agent;
+  const data = {
```
james-siteclick, line 10: I think we should probably be hydrating service models with data and then mapping/saving those back to the CRM.  This does involve lots of mapping, but hopefully handles some of the validation/business logic since those models already contain validators.

I think the InvoiceAccount model probably contains the full structure you need so I would aim to hydrate this and then persist it.

src/modules/companies/validators/invoice-accounts.js
```diff
@@ -0,0 +1,106 @@
+const Joi = require('@hapi/joi');
+
+const CONTACT_TYPES = ['person', 'department'];
+const COMPANY_TYPES = ['individual', 'limitedCompany', 'partnership', 'limitedLiabilityPartnership'];
```
james-siteclick, line 4: I think the majority of this validation logic probably lives in the service models.

Where only that property is affected, the validation logic can be in the setters/getters.

Where you have dependencies between properties (e.g. in the address model), we could have a:

`isValidAddress()`

method to check the fully populated model.

src/lib/mappers/company-address.js
```diff
@@ -0,0 +1,22 @@
...
+  const companyAddress = new CompanyAddress(row.companyAddressId);
+  companyAddress.fromHash({
+    dateRange: new DateRange(row.startDate, row.endDate),
+    companyId: row.companyId,
+    isDefault: row.isDefault
+  });
+
+  if (row.role) companyAddress.roleName = row.role.name;
+  if (row.address) companyAddress.address = addressMapper.crmToModel(row.address);
+  if (row.addressId) companyAddress.address = new Address(row.addressId);
```
james-siteclick, line 17: It may be possible that on the CRM end, when doing a bookshelf call withRelated models, that it responds with both `address` and `addressId` properties populated.
In this case I think the address from the line above would get overwritten - maybe we need:
```
if (row.address) {
  companyAddress.address = addressMapper.crmToModel(row.address);
}
else if (row.addressId) companyAddress.address = new Address(row.addressId);
```

src/lib/mappers/company-contact.js
```diff
@@ -3,26 +3,32 @@
...
     'dateUpdated'
   ]);
 
+  companyContact.dateRange = new DateRange(entity.startDate, entity.endDate);
+
   if (entity.role) {
     companyContact.role = contactRoleMapper.crmToModel(entity.role);
   }
 
   if (entity.contact) {
```
james-siteclick, line 24: Same here, I wonder if we might need:
```
if (entity.contact) {
  // ...
} else if (entity.contactId) {
  // ...
}
```

src/lib/models/contact-v2.js
```diff
@@ -1,7 +1,31 @@
...
+  : null;
+
+const contactPersonSchema = Joi.object({
+  title: Joi.string().allow(null).optional(),
+  firstName: Joi.string().required(),
+  middleInitials: Joi.string().allow(null).optional(),
+  lastName: Joi.string().required(),
+  suffix: Joi.string().allow(null).optional(),
+  department: Joi.string().allow(null).replace(/\./g, '').optional(),
+  dataSource: Joi.string().valid(Object.values(DATA_SOURCE_TYPES)).allow(null).optional()
```
james-siteclick, line 22: I think the dataSource property should probably not be nullable/optional

src/lib/services/invoice-account-addresses-service.js
```diff
@@ -0,0 +1,33 @@
...
+  const invoiceAccountAddress = new InvoiceAccountAddress();
+  return invoiceAccountAddress.fromHash({
+    dateRange: new DateRange(startDate, null),
+    address: addressModel,
+    agentCompany: agentCompanyModel,
+    contact: contactModel
+  });
+};
+
+const createInvoiceAccountAddress = async (invoiceAccount, startDate) => {
```
james-siteclick, line 15: I think this might be more flexible for the future if its signature was:
```
const createInvoiceAccountAddress = async (invoiceAccount, invoiceAccountAddress, startDate) => {
```

src/lib/services/invoice-account-addresses-service.js
```diff
@@ -0,0 +1,33 @@
...
+  const { address, agentCompany, contact } = invoiceAccount.invoiceAccountAddresses[0];
+  const data = {
+    addressId: address.id,
+    startDate,
+    agentCompanyId: agentCompany ? agentCompany.id : null,
+    contactId: contact ? contact.id : null
+  };
+
+  const invoiceAccountAddress = await invoiceAccountsConnector.createInvoiceAccountAddress(invoiceAccount.id, data);
+  invoiceAccount.invoiceAccountAddresses[0].id = invoiceAccountAddress.invoiceAccountAddressId;
```
james-siteclick, line 25: Although I have done this elsewhere myself, I think it would be better not to manipulate the arguments and return a completely fresh model, e.g.:
```
  const invoiceAccountAddress = await invoiceAccountsConnector.createInvoiceAccountAddress(invoiceAccount.id, data);
return mappers.crm.invoiceAccountAddress.crmToModel(invoiceAccountAddress);

```

src/lib/services/regions-service.js
```diff
@@ -0,0 +1,10 @@
+const regionsRepo = require('../connectors/repos/regions');
+const { NotFoundError } = require('../errors');
+
+const getRegionCode = async regionId => {
```
james-siteclick, line 4: It might be more flexible for the future to be able to get a Region service model - think the mapping code already exists:
```
const getRegion = async(regionId) => {
  const region = await regionsRepo.findOne(regionId);
  if (region) return mappers.region.dbToModel(region);
  throw new NotFoundError(`Region ${regionId} not found`);
}
```

The consuming code can be updated to pull out the region ID

src/modules/companies/controller.js
```diff
@@ -6,9 +6,26 @@ const companyContactsService = require('../../lib/services/company-contacts');
...
+const companiesService = require('../../lib/services/companies-service');
+const invoiceAccountsService = require('../../lib/services/invoice-accounts-service');
 const Boom = require('@hapi/boom');
+const { NotFoundError, InvalidEntityError } = require('../../lib/errors');
+const { envelope } = require('../../lib/response');
+
+// caters for error triggered in this service and 404s returned from the CRM
+const isNotFoundError = err => err instanceof NotFoundError || err.statusCode === 404;
+
+const mapErrorResponse = error => {
```
james-siteclick, line 18: Don't know if this `mapErrorResponse` code is similar to that in the billing module - if so we can make a shared one and put in src/lib

PR #1032, opened 2020-07-08
Feat: Adds resource for company contacts

swaff:
WATER-2721

Gets the company contacts by company id from the CRM V2 service.

Adds or promotes models and mappers to src/lib


src/lib/models/company-contact.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  get dateUpdated () { return this._dateUpdated; }
+  set dateUpdated (dateUpdated) {
+    this._dateUpdated = this.getDateTimeFromValue(dateUpdated);
+  }
+
+  get contact () { return this._contact; }
+  set contact (contact) {
+    if (contact !== null) {
```
james-siteclick, line 55: Is this check needed with the `assertIsNullableInstanceOf`?

src/lib/models/company-contact.js
```diff
@@ -0,0 +1,70 @@
...
+      validators.assertIsNullableInstanceOf(contact, Contact);
+    }
+    this._contact = contact;
+  }
+
+  get role () { return this._role; }
+  set role (role) {
+    if (role != null) {
```
james-siteclick, line 63: Is this check needed with the `assertIsNullableInstanceOf`?

src/lib/models/role-v2.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
```
james-siteclick, line 1: I wonder if we should rename this ContactRole and the other one DocumentRole if we are going to have different role models for the various jobs?
Alternatively the original Role may be flexible enough if it was sparsely populated?

src/modules/billing/services/contacts-service.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const contactsConnector = require('../../../lib/connectors/crm-v2/contacts');
-const mappers = require('../mappers');
```
james-siteclick, line 4: Is there a reason for changing this pattern?

swaff: The contacts mapper originally found in the billing module has been promoted up to the src/lib/mappers directory so it can be shared with the companies module.

PR #1040, opened 2020-07-10
Add API to search companies house

james-siteclick:
Adds an API at
/water/1.0/companies-house/search/companies?q=

To search for companies at Companies House.
Maps responses to service model shape.


src/modules/companies-house/routes.js
```diff
@@ -0,0 +1,21 @@
...
+const Joi = require('@hapi/joi');
+const controller = require('./controller');
+
+module.exports = {
+  getCompaniesHouseCompanies: {
+    path: '/water/1.0/companies-house/search/companies',
+    method: 'GET',
+    handler: controller.getCompaniesHouseCompanies,
+    config: {
+      description: 'Gets a list of returns for the specified company',
```
Stephan-Hough, line 12: Is this description correct? Should it be something like: 'Gets a list of companies for the specified name'

PR #1047, opened 2020-07-16
Feat: Adds a supplementary integration test

swaff:
WATER-2744

Updates the query to get supplementary charge version

 - Includes a filter to only return current charge versions

Adds supplementary tests

 - Create annual batch
 - Then create supplementary batch
 - Ensures the credit and debits are created
 - Ensures the transactions match locally and at the charge module


integration-tests/billing/services/data/charge-versions.js
```diff
@@ -1,8 +1,17 @@
...
+  endDate: null,
+  source: 'nald'
+};
+
+exports.cv2 = {
+  scheme: 'alcs',
+  versionNumber: 2,
+  startDate: '2019-04-01',
+  status: 'current',
+  endDate: '2020-01-01',
```
Stephan-Hough, line 15: I think the end date here should be set to null

PR #1063, opened 2020-07-27
Fix: Prevents transaction description being null

swaff:
WATER-2778

Updates the creation of transaction descriptions to use the purpose use
name if the charge element has no description.


src/lib/models/transaction.js
```diff
@@ -19,6 +19,10 @@ const statuses = {
   error: 'error'
 };
 
+const getDescriptionFromChargeElement = chargeElement => {
+  return chargeElement.description || chargeElement.purposeUse.name;
+};
+
 const getTwoPartTariffTransactionDescription = (transaction) => {
```
james-siteclick, line 26: I wonder if this function would need to change slightly to handle a charge element with no description - presumably it should either say:
  return `${prefix} part ${purpose} charge at ${description}`;
or 
  return `${prefix} part ${purpose} charge`;   // when description empty

PR #1066, opened 2020-07-28
Fix KPI data issues

Stephan-Hough:
This is to allow certain sections of the KPI dashboard to load even if some data objects are not available
--controller boom not found removed to return empty data objects instead
-- tests fixed
-- naming licences mapper for month fixed.


src/modules/kpi-reporting/controller.js
```diff
@@ -15,9 +15,8 @@ const getErrorMessage = (data, errorMessage) => {
  * @returns {Object} an object containing all the data sets required for the KPI UI
  */
 const getKPIData = async (request) => {
-  // get the last 2 returns cycles parameters
-  const returnCycles = await helpers.returns.date.createReturnCycles().slice(-2);
-  const [registrations, delegatedAccess, returnsDataMonthly, returnsDataCycle1, returnsDataCycle2, licenceNamesData] = await Promise.all([
+  const returnCycles = await getReturnCycles();
+  const [registrationsData, delegatedAccessData, returnsDataMonthly, returnsDataCycle1, returnsDataCycle2, licenceNamesData] = await Promise.all([
```
swaff, line 19: Would it be better to use `Promise.allSettled` here? This would allow one or more errors, but with all of the taks being tried. E.g. 3 out of 6 could fail leaving 3 resuts to display.

Using Promise.all would mean that the first failure causings the full set to be rejected.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

src/modules/kpi-reporting/lib/mappers.js
```diff
@@ -43,7 +43,7 @@ const mapLicenceNamesData = (data) => {
       acc.monthly.push(
         {
           ...row,
-          month: months[row.month],
+          month: months[row.month - 1],
```
timtamimi, line 46: Again, not a problem, just an advisory: It might be worth adding an inline comment here to explain the -1 (as I'm not sure why it's -1, but I'm sure there's a very good reason!) :)

timtamimi (approved):
Looks fine - Please note the two small comments I left in the review though. :)

test/modules/kpi-reporting/controller.js
```diff
@@ -56,10 +56,19 @@ experiment('./modules/kpi-reporting/controller', () => {
...
+          registrations: emptyResponse,
+          delegatedAccess: emptyResponse,
+          returnsMonthly: emptyResponse,
+          returnsCycle1: {},
+          returnsCycle2: {},
+          licenceNames: emptyResponse
+        }
+      };
+      const response = await controller.getKpiData({});
+      expect(response).to.be.equal(emptyDataRespomse);
```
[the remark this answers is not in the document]
Stephan-Hough: well spotted!

src/modules/kpi-reporting/lib/mappers.js
Stephan-Hough: good shout I added a comment -- it's because of the array starting with index 0.

PR #1087, opened 2020-08-11
Feat/water 2768 de minimis

james-siteclick:
Reads de-minimis flag from charge module customer financial year summary and maps to service models at Invoice level.
Reads de-minimis flag from charge module transactions and maps to service models at Transactions level.  Stores flag in water.billing_transactions.is_deminimis.
When running supplementary billing, does not credit transactions that were subject to de-minimis


test/modules/billing/mappers/totals.js
```diff
@@ -89,7 +89,7 @@ experiment('modules/billing/mappers/totals', () => {
 
     experiment('when the customer is present in the bill run summary', () => {
       beforeEach(async () => {
-        result = totalsMapper.chargeModuleBillRunToInvoiceModel(data.billRun, INVOICE_ACCOUNT_NUMBER, 2021);
+        result = totalsMapper.chargeModuleBillRunToInvoiceModel(data, INVOICE_ACCOUNT_NUMBER, 2021);
```
timtamimi, line 92: I doubt it's a big deal, but I feel like `2021` here should be an env variable

src/modules/billing/services/batch-service.js
```diff
@@ -150,28 +155,56 @@ const saveInvoicesToDB = async batch => {
...
+ * @param {Batch} batch
+ * @return {Promise}
+ */
+const persistTotals = batch => {
+  const changes = batch.totals.pick([
+    'invoiceCount',
+    'creditNoteCount',
+    'netTotal'
+  ]);
+  return newRepos.billingBatches.update(batch.id, changes, false);
```
DGajewska, line 184: I think you've got an extra param

src/modules/billing/services/transactions-service.js
```diff
@@ -79,19 +77,56 @@ const setErrorStatus = transactionId =>
...
+ * @return {Promise}
+ */
+const clearDeMinimisByTransactionIds = partialRight(updateDeMinimis, false);
+
+/**
+ * Sets the de-minimis flags by transaction IDs
+ * @param {Array<String>} externalIds
+ * @return {Promise}
+ */
+const setDeMinimisByTransactionIds = partialRight(updateDeMinimis, true);
```
DGajewska, line 95: I don't think these functions are used anywhere. Did you mean to use them below?

src/modules/billing/services/batch-service.js
james-siteclick: Thanks, this should be fixed now

src/modules/billing/services/transactions-service.js
james-siteclick: Thanks, this should be fixed now

PR #1093, opened 2020-08-14
Feat: minimum charge

DGajewska:
WATER-2767

* Adds minimum charge related flags to db tables
* Adds flags to models and mappers
* Adds isMinimumCharge flag to transaction processor


src/modules/billing/mappers/api/invoice.js
```diff
@@ -9,17 +10,26 @@ const containsWaterUndertaker = invoiceLicences => {
...
   'invoiceAccount.company.name': 'name',
   'totals.netTotal': 'netTotal',
   'invoiceLicences[].licence.licenceNumber': 'licenceNumbers[]',
   'financialYear.yearEnding': 'financialYearEnding',
-  invoiceLicences: {
+  invoiceLicences: [{
     key: 'isWaterUndertaker',
     transform: containsWaterUndertaker
+  }, {
+    key: 'minimumChargeApplies',
```
james-siteclick, line 29: We generally have a convention of starting the property name with `is` for a boolean e.g. `isMinimumChargeApplied`

src/modules/billing/services/charge-processor-service/transactions-processor.js
```diff
@@ -99,6 +101,20 @@ const isCompensationChargesNeeded = (batch, chargeVersion) => {
...
+/**
+ * Predicate to check whether the minimum charge applies
+ * @param {DateRange} chargePeriod of the charge element
+ * @param {ChargeVersion} chargeVersion
+ * @return {Boolean}
+ */
+const doesMinimumChargeApply = (chargePeriod, chargeVersion) => {
+  const { dateRange, changeReason } = chargeVersion;
+  const chargeVersionStartDate = moment(dateRange.startDate);
+  const isSharedStartDate = moment(chargePeriod.startDate).isSame(chargeVersionStartDate);
```
james-siteclick, line 113: Just to guard against any funny timestamp issues:
```
const isSharedStartDate = moment(chargePeriod.startDate).isSame(chargeVersionStartDate, 'day');
```

src/modules/billing/services/charge-processor-service/transactions-processor.js
```diff
@@ -99,6 +101,20 @@ const isCompensationChargesNeeded = (batch, chargeVersion) => {
...
+  if (isNull(changeReason)) return false;
```
james-siteclick, line 114: Could maybe simplify as:
```
return isSharedStartDate && get(changeReason, 'triggersMinimimCharge', false)
```

src/lib/models/invoice.js
```diff
@@ -187,6 +196,12 @@ class Invoice extends Model {
   get isDeMinimis () {
     return this._isDeMinimis;
   }
+
+  toJSON () {
```
james-siteclick, line 200: This `toJSON` method doesn't appear to add anything to the functionality of the super class, so could probably be safely removed

src/modules/billing/mappers/api/invoice.js
```diff
@@ -16,10 +16,10 @@ const map = {
   'totals.netTotal': 'netTotal',
   'invoiceLicences[].licence.licenceNumber': 'licenceNumbers[]',
   'financialYear.yearEnding': 'financialYearEnding',
-  invoiceLicences: {
```
james-siteclick, line 19: I think this code was working because it divides the invoices into water companies/other abstractors in the UI:
```
  invoiceLicences: {
    key: 'isWaterUndertaker',
    transform: containsWaterUndertaker
  }
```

src/modules/billing/mappers/api/invoice.js
```diff
@@ -28,7 +28,9 @@ const map = {
  * @return {Object}
  */
 const modelToBatchInvoice = invoice => {
-  return objectMapper(invoice, map);
+  const mappedInvoice = objectMapper(invoice, map);
+  objectMapper.setKeyValue(mappedInvoice, 'isMinimumChargeApplied', invoice.isMinimumChargeApplied());
```
james-siteclick, line 32: Is this property needed in the UI list view?
If so, could this be added to the `map` above?

src/modules/billing/mappers/charge-module-decorators.js
```diff
@@ -47,16 +59,22 @@ const decorateInvoice = (invoice, cmResponse) => {
...
-  const transactionMap = indexChargeModuleTransactions(customerFinYearSummary.transactions);
+  const transactionsByLicence = indexTransactionsByLicenceNumber(customerFinYearSummary.transactions);
+  // Decorate each invoice transaction and add missing transactions created in the CM
   invoice.invoiceLicences.map(invoiceLicence => {
-    invoiceLicence.transactions.map(transaction => decorateTransaction(transaction, transactionMap));
+    transactionsByLicence.get(invoiceLicence.licence.licenceNumber).map(cmTransaction => {
+      const serviceTransaction = invoiceLicence.transactions.find(trans => trans.externalId === cmTransaction.id);
+      if (serviceTransaction) {
+        return decorateTransaction(serviceTransaction, cmTransaction);
+      }
```
james-siteclick, line 75: I wonder if we should throw an error if here if:  `!cmTransaction.minimumChargeAdjustment ` as that is the only type of transaction we would be expecting.

james-siteclick (changes requested):

src/modules/billing/mappers/api/invoice.js
DGajewska: Apologies, didn't realise I'd changed it. Thank you for pointing it out :)

src/lib/models/invoice.js
DGajewska: Removed

src/modules/billing/mappers/api/invoice.js
DGajewska: I was unable to accomplish it in the map, however, I no longer use it in the UI, so I have removed it

src/modules/billing/mappers/charge-module-decorators.js
DGajewska: Yes, good call!

james-siteclick (approved):

PR #1097, opened 2020-08-17
Feat: Charge Versions APIs

swaff:
WATER-2747

- Promotes charge element and version models and mappers to lib
- Updates the charge version APIs to use the model objects

This will change the display of data in the UI which is fixed in https://github.com/DEFRA/water-abstraction-ui/pull/1367


src/lib/models/date-range.js
```diff
@@ -106,6 +106,16 @@ class DateRange {
     return new DateRange(startDate, endDate);
   }
 
+  /**
+   * Takes an input and turns it into a string representation of a date
+   * in the format YYYY-MM-DD
+   *
+   * @param {*} input A date, moment, string or null to format
+   */
+  static formatDate (input) {
```
james-siteclick, line 115: I'm not certain if this should live here?  Would it be better in src/lib/date or similar?

PR #1124, opened 2020-09-01
WATER-2786 - Added GET /water/1.0/licences/licence-accounts

timtamimi:
Original ticket for context: https://eaflood.atlassian.net/browse/WATER-2786

This adds an endpoint that proxies the two newly added CRMv2 Routes for fetching information about the responsible company for a licence, and the invoice accounts associated with that licence.

### DIFF
--- a/src/lib/connectors/crm-v2/documents.js
+++ b/src/lib/connectors/crm-v2/documents.js
@@ -2,12 +2,17 @@
 
 const urlJoin = require('url-join');
 const Joi = require('joi');
+const moment = require('moment');
 
 const { serviceRequest } = require('@envage/water-abstraction-helpers');
 const config = require('../../../../config');
 
 const VALID_LICENCE_NUMBER = Joi.string().required().example('01/123/R01');
 const VALID_GUID = Joi.string().guid().required();
+
+const getDocumentUrl = (...tail) => {
+  return urlJoin(config.services.crm_v2, 'document', ...tail);
+};
 
 const getDocumentsUrl = (...tail) => {
   return urlJoin(config.services.crm_v2, 'documents', ...tail);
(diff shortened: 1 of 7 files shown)


src/lib/connectors/crm-v2/documents.js
```diff
@@ -2,13 +2,18 @@
...
 const { serviceRequest } = require('@envage/water-abstraction-helpers');
 const config = require('../../../../config');
 
 const VALID_LICENCE_NUMBER = Joi.string().required().example('01/123/R01');
 const VALID_GUID = Joi.string().guid().required();
 
+const getDocumentUrl = (...tail) => {
+  return urlJoin(config.services.crm_v2, 'document', ...tail);
+};
+
```
Stephan-Hough, line 16: The code matches the details in the ticket but I think we should re-use the existing documents connector code because we are working with the same entity:

const getDocumentsUrl = (...tail) => {
  return urlJoin(config.services.crm_v2, 'documents', ...tail);
};

This will require a change to the CRM API endpoint to match and will add a comment to that pull request.

src/lib/connectors/crm-v2/documents.js
```diff
@@ -57,7 +62,22 @@ const getDocumentRole = documentRoleId => {
   return serviceRequest.get(getDocumentRolesUrl(documentRoleId));
 };
 
+/**
+ * Fetch a document by documentRef and date, to identify the responsible company
+ *  @param {String} documentRef
+ *  @param {String} date
+ */
+const getDocumentByRefAndDate = async (documentRef, date) => {
+  return serviceRequest.get(getDocumentUrl('search'), {
```
Stephan-Hough, line 71: If we agree on the above comment then this line changes to: 
return serviceRequest.get(getDocumentsUrl('search'), {

src/lib/services/licences.js
```diff
@@ -108,13 +110,29 @@ const updateIncludeInSupplementaryBillingStatusForSentBatch = async batchId => {
...
+ */
+const getLicenceAccountsByRefAndDate = async (documentRef, date) => {
+  //  First, fetch the company ID from the CRM
+  const { companyId } = await crmDocsConnector.getDocumentByRefAndDate(documentRef, date);
+
+  //  Secondly, take the company ID and fetch the invoice accounts for that company from the CRM
+  const invoiceAccounts = await crmCompaniesConnector.getInvoiceAccountsByCompanyId(companyId);
+
+  //  Return the invoice accounts
+  return invoiceAccounts;
```
Stephan-Hough, line 126: The invoice accounts should be mapped to the water service models before returning them to the UI and if there are no invoice accounts then return an empty array. The mappers and models already exist in ./src/lib i.e. 

return invoiceAccounts ? InvoiceAccountsMapper.crmToModel(invoiceAccounts) : [];

Stephan-Hough (changes requested):

src/lib/services/licences.js
timtamimi: Will do, but I think I'll need that line to be:

`return invoiceAccounts ? invoiceAccounts.map(invoiceAccount => invoiceAccountsMapper.crmToModel(invoiceAccount)) : [];`

This is in order to get the invoice accounts to map to the model one at a time (as it's returning an array of them)

I hope that's OK but if it's not let me know and I'll undo.

Stephan-Hough (approved):

PR #1126, opened 2020-09-02
WATER 2793 - Create charge test data for Acceptance tests

Stephan-Hough:
feat/water-2793
-- call to integration tests to create scenario from acceptance test set-up


timtamimi (approved):
Code seems sensible. I've left one syntax-related comment but it's no drama if you would rather ignore that.

PR #1128, opened 2020-09-03
Feat/water 2836 charge information 4 search existing companies

timtamimi:
Original ticket for context: https://eaflood.atlassian.net/browse/WATER-2836


Stephan-Hough (commented):
I think you need to add a test for the new companies route and another for the companies service.

PR #1140, opened 2020-09-14
Fix issue with large (e.g. annual) bill run

james-siteclick:
WATER-2748
* Fixes issue where large bill run did not publish enough jobs to process all transactions
* Moves setting of batch error status to job handler to ensure it is in error state as fast as possible
* Adds batch status checks to all on-complete handlers to ensure process will not proceed when batch errors
* Fixes mapping issue with missing invoice account address


src/modules/billing/jobs/create-charge-complete.js
```diff
@@ -93,29 +98,20 @@ const finaliseReadyBatch = async (job, messageQueue) => {
...
     await actions[status](job, messageQueue);
   } catch (err) {
     batchJob.logOnCompleteError(job);
-    await batchService.setErrorStatus(batchId, BATCH_ERROR_CODE.failedToCreateCharge);
     throw err;
   }
 };
 
-module.exports = handleCreateChargeComplete;
+module.exports = partialRight(createOnCompleteHandler, handleCreateChargeComplete, BATCH_STATUS.processing);
```
timtamimi, line 116: note to future self: Find out more about `partialRight()`. It seems useful.

timtamimi (approved):
With my limited knowledge, I think this all looks good. :) 

Only suggestion (which can be ignored) is that it might be worth putting `batchService.setErrorStatus` inside of `batchJob.logHandlingError` just for the sake of brevity.

src/modules/billing/jobs/create-charge-complete.js
james-siteclick: > Only suggestion (which can be ignored) is that it might be worth putting batchService.setErrorStatus inside of batchJob.logHandlingError just for the sake of brevity.

Done now thanks

PR #1141, opened 2020-09-15
Creates integration tests for scenario SB2

DGajewska:
WATER-2797

*Reorganises tests by batch type and scenario reference


james-siteclick (changes requested):
I think we should make the data in the scenarios match the spreadsheet as closely as possible.

james-siteclick (approved):

PR #1147, opened 2020-09-22
Feat: Delete licence agreement

DGajewska:
WATER-2784

### DIFF
--- a/src/modules/licences/controllers/agreements.js
+++ b/src/modules/licences/controllers/agreements.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const eventHelper = require('../lib/event-helper');
+const mapErrorResponse = require('../../../lib/map-error-response');
 const Boom = require('@hapi/boom');
 
 const licencesService = require('../../../lib/services/licences');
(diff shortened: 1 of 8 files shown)


src/modules/licences/routes/agreements.js
```diff
@@ -29,5 +29,22 @@ module.exports = {
...
+    method: 'DELETE',
+    path: '/water/1.0/agreements/{agreementId}',
+    handler: controller.deleteAgreement,
+    config: {
+      description: 'Deletes the agreement with the specified id',
+      validate: {
+        params: {
+          agreementId: Joi.string().uuid().required()
+        },
+        headers: async values => {
```
james-siteclick, line 44: Due to some recent tweaks I made to the internalCallingUser plugin, it's fairly easy now to restrict access for a route if needed to only allow users with a particular scope (similar to the UI project)
https://github.com/DEFRA/water-abstraction-service/blob/develop/src/modules/charge-versions/routes/charge-version-workflows.js#L22

src/modules/licences/controllers/agreements.js
```diff
@@ -19,5 +21,36 @@ const getLicenceAgreements = async request => {
...
+
+    if (!licenceAgreement) {
+      return Boom.notFound(`Agreement ${agreementId} not found`);
+    }
+
+    // Delete the licence agreement
+    await licencesService.deleteLicenceAgreementById(agreementId);
+
+    // log deletion in event log
+    await eventHelper.saveEvent(
```
james-siteclick, line 39: I wonder if the event logging code could also go in the service - however I'm happy to move this over later.

james-siteclick (approved):

PR #1148, opened 2020-09-22
Feat/water 2688 agreements api

james-siteclick:
WATER-2688
* Adds migrations to use a guid ID for water.financial_agreement_types table
* Tweaks model and service code to handle changes
* Adds POST endpoint to create a new licence agreement
* Protects all agreements routes using new IDM roles for agreements


src/lib/models/licence-agreement.js
```diff
@@ -34,10 +45,12 @@ class LicenceAgreement extends Model {
     this._agreement = agreement;
   }
 
-  get dateSigned () { return this._dateSigned; }
+  get dateSigned () {
+    return this._dateSigned ? this._dateSigned.format(DATE_FORMAT) : this._dateSigned;
```
timtamimi, line 49: I think if _dateSigned is null or undefined, this will return null or undefined. Is that intentinonal?

If this is intentional: Ignore me.

If this isn't intentional: I would suggest changing the `else` value to some sort of default fallback. E.g. Today's date or something.

src/lib/services/licence-agreements.js
```diff
@@ -0,0 +1,144 @@
...
+    // Log event and flag licence for supplementary billing
+    await Promise.all([
+      createEvent(EVENT_TYPES.create, licenceAgreement, issuer),
+      licencesService.flagForSupplementaryBilling(licence.id)
+    ]);
+
+    // Return the updated model
+    return licenceAgreement;
+  } catch (err) {
+    if (err.code === '23505') {
```
timtamimi, line 134: Neat.

src/plugins/internal-calling-user.js
```diff
@@ -5,6 +5,7 @@ const { get, set } = require('lodash');
 
 const { logger } = require('../logger');
 const idmConnector = require('../lib/connectors/idm');
+const User = require('../lib/models/user');
 
 const validateUserIsInternal = user => {
   if (user.application !== 'water_admin') {
```
timtamimi, line 11: I think I'm not 100% sure what this is doing. Might it be worth adding a comment?

It looks like it's a function to check whether the user is internal/admin, but the if statement is then checking if the user is NOT admin. Bit confused, but I think it's likely that I'm just not reading it correctly :)

timtamimi (commented):
All fine, but I (as usual) added a couple of queries. 

Grateful if you could cast your eyes on them. If you're satisfied I'll give it a ceremonial ✅  :)

src/lib/models/licence-agreement.js
james-siteclick: This is intentional

timtamimi (approved):

PR #1163, opened 2020-09-30
Fix bill run issues due to CM statuses

james-siteclick:
WATER-2870

Fixes issues arising because the CM now responds with an incomplete payload depending on bill run status.

* Wraps the CM API connector code in a method which will retry up to 10 times with exponential back-off
* Does not move batch from 'processing' to 'ready' until the 'refreshTotals' job has completed
* Puts the batch back into 'processing' status while a customer is removed

### DIFF
--- a/src/lib/connectors/charge-module/bill-runs-with-retry.js
+++ b/src/lib/connectors/charge-module/bill-runs-with-retry.js
@@ -29,7 +29,8 @@ const getOrThrowIfGeneratingSummary = async (method, ...args) => {
 };
 
 /**
- *
+ * Wraps a charge module API method so that it retries a number
+ * of times, awaiting a status other than 'generating_summary'
  * @param {String} method - charge module bill run connector method
  * @param  {...any} args - additional arguments
  */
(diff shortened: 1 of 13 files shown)


src/lib/connectors/charge-module/bill-runs-with-retry.js
```diff
@@ -0,0 +1,49 @@
...
+
+const { get, partial } = require('lodash');
+const promiseRetry = require('promise-retry');
+const { logger } = require('../../../logger');
+
+const billRunsApi = require('./bill-runs');
+
+const isCMGeneratingSummary = cmResponse => get(cmResponse, 'billRun.status') === 'generating_summary';
+
+const options = {
```
timtamimi, line 11: I'm wondering if those options should sit in the top-level config object at the root of the application.

Not important - just a thought.

timtamimi (approved):
Noice. Looks good. I left one suggestion which can be ignored at your discretion.

Another thing that crossed my mind was:
`Is there much point to having two charge module connectors? One for attempting a call with retries, and one without?` - but that can be a battle for a different day. :)

src/lib/connectors/charge-module/bill-runs-with-retry.js
james-siteclick: I have moved to the config to the app-level config object as requested.

PR #1174, opened 2020-10-06
Feat/water 2856 paper forms api

james-siteclick:
Adds an API to find the returns that are due for a list of licences.
The returns are grouped by CRM v2 document date range, and the returns/licence contact are also returned for each group.


src/modules/returns/routes/routes.js
```diff
@@ -0,0 +1,65 @@
...
+
+  getIncompleteReturns: {
+    path: '/water/1.0/returns/incomplete',
+    method: 'GET',
+    handler: controller.getIncompleteReturns,
+    config: {
+      description: 'Gets a list of incomplete returns by licence number - supports paper forms flow',
+      validate: {
+        query: Joi.object({
+          licenceNumbers: Joi.array().single().min(1).items(Joi.string().required()).required()
```
timtamimi, line 59: This Joi definition is a little difficult to read. Anything we can do to simplify it?

No drama if not. It's not a blocker.

timtamimi (approved):

PR #1175, opened 2020-10-06
Chore: Promote PDF generation code to the lib dir

swaff:
WATER-2858

Moves the PDF and HTML (for PDF) generation code to lib to allow the
service to be consumed more easily from the modules.

 - src/lib/services/pdf-generation/html
   - Adds `createHtmlFromTemplate`
   - Adds `createHtmlFromScheduledNotification`
 - src/lib/services/pdf-generation/pdf
   - Adds `createPdf`
   - Adds `createPdfFromScheduledNotification`
   - Adds `createPdfFromTemplate`


src/lib/models/scheduled-notification.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+const Model = require('./model');
+
+const validators = require('./validators');
+
+class ScheduledNotification extends Model {
+  get recipient () { return this._recipient; }
+  set recipient (value) {
+    validators.assertNullableString(value);
```
james-siteclick, line 10: I think the recipient here is either null or an email address - would we wish to add validation for email here?

james-siteclick (changes requested):

src/lib/models/scheduled-notification.js
swaff: I don't think so without a wider change to the notifications. I found entries in the database where the recipient is 'n/a' so think it might be risky to enforce the email validation here.

james-siteclick (approved):

PR #1187, opened 2020-10-16
Adds a route for reinstating a disabled user.

timtamimi:
The route can be found at `POST /water/1.0/user/internal/{id}/reinstate`

Paired with https://github.com/DEFRA/water-abstraction-ui/pull/1434


DGajewska (approved):
Looks good, but I'm wondering whether these endpoints should be updated to validate (in the routes) and use the `defra-internal-user-id` header (in the controller) instead of making a call to get the calling user data, as we do now

timtamimi:
Spoken to UCD about this and a decision was made against this change, so I'm closing this PR.

PR #1189, opened 2020-10-16
Feat: Get licences without charge versions

swaff:
WATER-2888

Adds the start date for fetching licences without charge version to the
config file.

Adds SQL query to find all licences that have no charge versions where
the licence version is current or superseded.

Adds new route, controller handler and service code to combine licences
with licence holder role.

### DIFF
--- a/src/modules/licences/controllers/licences.js
+++ b/src/modules/licences/controllers/licences.js
@@ -38,8 +38,24 @@ const getValidLicenceDocumentByDate = async request => {
   return documentsService.getValidDocumentOnDate(licence.licenceNumber, date);
 };
 
+const getLicencesWithoutChargeVersions = async () => {
+  const licences = await licencesService.getLicencesWithoutChargeVersions();
+
+  const promises = licences.map(licence => {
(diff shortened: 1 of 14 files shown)


src/modules/licences/controllers/licences.js
```diff
@@ -38,8 +38,24 @@ const getValidLicenceDocumentByDate = async request => {
   return documentsService.getValidDocumentOnDate(licence.licenceNumber, date);
 };
 
+const getLicencesWithoutChargeVersions = async () => {
+  const licences = await licencesService.getLicencesWithoutChargeVersions();
+
+  const promises = licences.map(licence => {
+    return documentsService.getValidDocumentOnDate(licence.licenceNumber, licence.startDate)
```
timtamimi, line 45: Small issue. This fails if a licence doesn't have any corresponding document records. 

The API then returns:
```
message: Current or superseded document not found for AN/xxx/xxxx/xxx on 2020-03-13, stack: NotFoundError: Current or superseded document not found for AN/xxx/xxxx/xxx on 2020-03-13
```

timtamimi (changes requested):

timtamimi (approved):

PR #1196, opened 2020-10-21
Bugfix/companies house api proxy failure

timtamimi:
This PR does two things:
1. Changes the change-reasons API endpoint to return the response in an object instead of an array
2. Moves the Companies House service method to use request-promise instead of the helper.


timtamimi:
This PR *will* reduce test coverage a bit. This is because the Companies House service is now being called by request-promise which is very hard to mock in tests.

james-siteclick (approved):

PR #1198, opened 2020-10-22
Removes error throwing

timtamimi:


src/modules/charge-versions/services/charge-version-workflows.js
```diff
@@ -40,17 +39,14 @@ const getAll = () => service.findAll(chargeVersionWorkflowsRepo.findAll, chargeV
  */
 const getLicenceHolderRole = async chargeVersionWorkflow => {
   const { licenceNumber } = chargeVersionWorkflow.licence;
-  const { startDate } = chargeVersionWorkflow.chargeVersion.dateRange;
+  const { startDate } = chargeVersionWorkflow.chargeVersion.dateRange || { chargeVersion: { startDate: null } };
```
swaff, line 42: Is the default object missing the dateRange object?

I wonder if the following is more readable?

const startDate  = get(chargeVersionWorkflow, 'chargeVersion.dateRange.startDate', null)

timtamimi:
Closed this PR in favour for a new approach

PR #1201, opened 2020-10-23
Feat/water 2885 charge version workflows

timtamimi:
Ticket: https://eaflood.atlassian.net/browse/WATER-2885

Linked to: https://github.com/DEFRA/water-abstraction-ui/pull/1455


src/modules/charge-versions/controllers/charge-version-workflow.js
```diff
@@ -41,13 +42,18 @@ const postChargeVersionWorkflow = async request => {
   const { licenceId } = request.payload;
   const { chargeVersion, user } = request.pre;
 
+  //  Create a tweaked version of the charge version object
+  //  This is because the 'status' at this point would not validate against the charge version model
+  const parsedCV = chargeVersion;
+  parsedCV.status = 'draft';
```
Stephan-Hough, line 48: do you have to declare a new variable or probably easier to simply add the status or update it i.e. 
chargeVersion.status = 'draft';

src/modules/charge-versions/services/charge-version-workflows.js
```diff
@@ -40,16 +39,14 @@ const getAll = () => service.findAll(chargeVersionWorkflowsRepo.findAll, chargeV
  */
 const getLicenceHolderRole = async chargeVersionWorkflow => {
   const { licenceNumber } = chargeVersionWorkflow.licence;
-  const { startDate } = chargeVersionWorkflow.chargeVersion.dateRange;
+  const startDate = get(chargeVersionWorkflow, 'chargeVersion.dateRange.startDate', null);
```
Stephan-Hough, line 42: we shouldn't ever have a situation where the charge version start date is null. If this is the case then the UI is at fault

src/modules/charge-versions/services/charge-version-workflows.js
```diff
@@ -40,16 +39,14 @@ const getAll = () => service.findAll(chargeVersionWorkflowsRepo.findAll, chargeV
...
   const doc = await documentsService.getValidDocumentOnDate(licenceNumber, startDate);
 
-  if (!doc) {
-    throw new errors.NotFoundError(`Current or superseded document not found for ${licenceNumber} on ${startDate}`);
-  }
+  const role = doc ? doc.getRoleOnDate(Role.ROLE_NAMES.licenceHolder, startDate) : {};
```
Stephan-Hough, line 45: Is it OK to pass an empty object if the document role doesn't exist?

src/modules/charge-versions/services/charge-version-workflows.js
timtamimi: You're right - it **shouldn't** happen (ideally)

The `null` is a default value/safety net. Would you prefer that we removed this such that the UI throws an error if any of the objects have a null start date?

src/modules/charge-versions/controllers/charge-version-workflow.js
timtamimi: You're right. It's not needed. I've changed this.

src/modules/charge-versions/services/charge-version-workflows.js
Stephan-Hough: Yes I think it should throw and error otherwise we might end up with incomplete charge info

Stephan-Hough (approved):

PR #1207, opened 2020-10-29
Feat/Edit charge versions - licence id added to charge versions

Stephan-Hough:
Feat/water 2806
-- licence_id added to water.charge_versions
-- new route to get charge versions by licence id
-- getChargeVerionById updated.
-- charge version mapper updated


DGajewska (approved):

migrations/sqls/20201026081506-add-charge-version-licence-id-up.sql
```diff
@@ -0,0 +1,9 @@
+/* Add the licence id to charge versions */
+ALTER TABLE water.charge_versions
+  ADD COLUMN IF NOT EXISTS "licence_id" uuid;
```
DGajewska, line 3: Should this have a `NOT NULL` constraint?

DGajewska: If so, this will have knock on effects with the charge version workflow in the UI and the `postCreateFromWorkflow` endpoint in the `charge-versions` module

Stephan-Hough: I had to leave the not null out because I noticed there are some charge versions that don't have a licence ref. This means the licence id can not be populated. But more importantly the import feature will fail because it copies the charge version data from NALD and then it updates the licence id. 

The import SQL statement could be changed but this is already so complex I didn't want to add another join to it only to get the licence id. When we switch the charging import feature off we could add the constraint and at the same time remove the licence ref as this is not really needed if we will use the licence id instead.

DGajewska (approved):

PR #1216, opened 2020-11-05
Round billing volumes to 6dp not 3dp

james-siteclick:
WATER-2926

Fixes an issue where charge elements with annual volumes with 4-6 decimal points of precision could not be processed as the billing volume rounded to 3dp exceeded the annual volume.

Uses the decimal-light JS library for calculations during TPT matching.


timtamimi (approved):

src/lib/mappers/user.js
```diff
@@ -15,7 +16,7 @@ const dbToModel = data => {
 
 const pojoToModel = dbToModel;
 
-const modelToDb = model => model.toJSON();
+const modelToDb = model => isObject(model) ? model.toJSON() : null;
```
timtamimi, line 19: I'm wondering if the `:null` on the end should be `:{}` (So that the returned type is always an object)

If the null is intentional, ignore this comment.

src/lib/models/billing-volume.js
```diff
@@ -79,15 +80,22 @@ class BillingVolume extends Model {
...
 
   set calculatedVolume (calculatedVolume) {
-    validators.assertNullableQuantity(calculatedVolume);
-    this._calculatedVolume = isNull(calculatedVolume) ? null : parseFloat(calculatedVolume);
+    if (isNull(calculatedVolume)) {
+      this._calculatedVolume = null;
+    } else {
+      const value = new Decimal(calculatedVolume);
+      if (value.isNegative()) {
+        throw new Error(`Expected zero or positive number or decimal, got ${value.toNumber()}`);
```
timtamimi, line 95: Not important. Just nitpicking 👇🏽 

<pedantry>
Consider rewording the error to `Expected zero or positive figure` purely because a number can be both positive and decimal.
</pedantry>

src/lib/models/billing-volume.js
```diff
@@ -172,41 +180,67 @@ class BillingVolume extends Model {
...
   }
+
+  /**
+   * Gets either the approved volume, or the calculated volume if the billing
+   * volume is not yet approved, as a decimal
+   * @return {Decimal}
+   */
+  get approvedOrCalculatedVolume () {
+    if (this.isApproved) {
+      return isNull(this.volume) ? null : new Decimal(this.volume);
```
timtamimi, line 233: Wondering if the `null` should be a zero instead? Ignore if not.

timtamimi (approved):
Looks good. I left three ignorable comments/queries.

src/lib/mappers/user.js
james-siteclick: This one is intentional - it gets mapped to DB column which is either a jsonb containing the user or null

src/lib/models/billing-volume.js
james-siteclick: Ah-ha, I meant Decimal instance, I've re-worded the message to hopefully make it clearer

src/lib/models/billing-volume.js
james-siteclick: Null is OK here

PR #1219, opened 2020-11-09
Feat: User can only run 1 annual or TPT bill run per region, FY and season

DGajewska:
WATER-2822

* Updates batch service to check for sent bill runs for type, region, FY and season
* Refactors batch repo calls for sent batches


src/lib/connectors/repos/billing-batches.js
```diff
@@ -113,40 +113,38 @@ const create = async data => {
...
+    'billingInvoices',
+    'billingInvoices.billingInvoiceLicences',
+    'billingInvoices.billingInvoiceLicences.licence'
+  ];
+
+  return findSentBatchesForRegion(regionId, filters, withRelated);
 };
 
+const findSentTPTBatchesForRegion = async regionId =>
+  findSentBatchesForRegion(regionId, { batch_type: BATCH_TYPE.twoPartTariff });
```
james-siteclick, line 146: I wonder if the methods in this file could be simpler by making use of the `findMany` repo helper

src/modules/billing/services/batch-service.js
```diff
@@ -55,6 +56,34 @@ const getMostRecentLiveBatchByRegion = async regionId => {
...
+ * Checks if there is:
+ *  - an existing batch in flight for the region
+ *  - a given batch for the same type, region, year and season (excluding supplementary)
+ * @param {String} regionId
+ * @param {String} batchType
+ * @param {Number} toFinancialYearEnding
+ * @param {Boolean} isSummer
+ * @return {Batch|null} if batch exists
+ */
+const getExistingOrDuplicateSentBatch = async (regionId, batchType, toFinancialYearEnding, isSummer) => {
```
james-siteclick, line 78: At the moment it seems we are doing 2 separate queries and then some additional filtering to find the duplicate/sent batches.

Could this be simplified by:
* A query finds all batches in region
* Putting them through a single `find()` which looks for either an open batch in the same region, or a sent batch of the same type/region/financial year/season (if the requested new batch is annual/two_part_tariff)

I think we may then need fewer repo methods too

james-siteclick (changes requested):

james-siteclick (approved):

PR #1221, opened 2020-11-10
Adds API endpoint in non-prod to delete all billing data

james-siteclick:
* Converts external_id field in water.charge_versions table to use unique varchar e.g regionCode:licenceId:versionNumber 
* Converts external_id field in water.charge_elements table to use unique varchar, e.g. regionCode:id
* Adds endpoint to truncate all billing/charge version tables and call import module to reschedule charge version import

### DIFF
--- a/src/lib/connectors/repos/queries/billing-batches.js
+++ b/src/lib/connectors/repos/queries/billing-batches.js
@@ -1,0 +1,16 @@
+'use strict';
+
+const deleteAllBillingData = `
+truncate 
+water.billing_batch_charge_versions, 
+water.billing_batch_charge_version_years, 
+water.billing_batches, 
+water.billing_invoice_licences,
+water.billing_invoices, 
+water.billing_transactions, 
+water.billing_volumes,
+water.charge_versions,
+water.charge_elements
+`;
+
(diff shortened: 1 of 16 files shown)


src/lib/connectors/repos/queries/billing-batches.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const deleteAllBillingData = `
+truncate 
+water.billing_batch_charge_versions,
```
DGajewska, line 5: This table no longer exists as the migrations removing it have now been merged

src/lib/services/charge-versions.js
```diff
@@ -175,8 +175,15 @@ const create = async chargeVersion => {
   return getByChargeVersionId(id);
 };
 
+/**
+ * Deletes all billing data in service (!)
+ * @return {Promise}
+ */
+const deleteAllChargeVersionData = () => chargeVersionRepo.deleteAllChargeVersionData();
```
DGajewska, line 182: It doesn't look like you call this anywhere

DGajewska (changes requested):

DGajewska (approved):

PR #1228, opened 2020-11-16
Feat: Exclude licences with charge version workflow in progress from bill runs

DGajewska:
WATER-2917

* Updates query which selects charge versions for billing
* Flags licence for supplementary billing when a charge version workflow is deleted
* Updates licence service to use correct method for flagging a licence for supplementary billing


src/modules/charge-versions/controllers/charge-version-workflow.js
```diff
@@ -82,8 +82,16 @@ const patchChargeVersionWorkflow = async (request, h) => {
   }
 };
 
-const deleteChargeVersionWorkflow = (request, h) =>
-  controller.deleteEntity(chargeVersionsWorkflowService.delete, h, request.params.chargeVersionWorkflowId);
+const deleteChargeVersionWorkflow = async (request, h) => {
+  const { chargeVersionWorkflow } = request.pre;
+  try {
+    await chargeVersionsWorkflowService.delete(chargeVersionWorkflow.id);
+    await licencesService.flagForSupplementaryBilling(chargeVersionWorkflow.licence.id);
```
james-siteclick, line 89: If we always need to flag a licence whenever we delete a charge version workflow, I'd be tempted to move this into the chargeVersionWorkflowService

src/modules/charge-versions/controllers/pre-handlers.js
```diff
@@ -46,5 +47,17 @@ const mapInternalCallingUser = (request, h) => {
   return mapOrThrowBoom('user', internalCallingUser, userMapper);
 };
 
+const loadChargeVersionWorkflow = async request => {
+  const { chargeVersionWorkflowId } = request.params;
+  const chargeVersionWorkflow = await chargeVersionWorkflowService.getById(chargeVersionWorkflowId);
+
+  if (!chargeVersionWorkflow) {
+    throw Boom.notFound(`No charge version workflow found with id: ${chargeVersionWorkflowId}`);
```
james-siteclick, line 55: I think you can return rather than throw a Boom error from a hapi pre-handler - this should halt the chain of processing and return the expected status code etc.

james-siteclick (changes requested):

james-siteclick (approved):

timtamimi (approved):

PR #1230, opened 2020-11-17
Foreign key and constraint added to charge versions table for licence id

Stephan-Hough:
feat/water-2806
-- added not null constraint and foreign key for charge versions


migrations/sqls/20201117091026-charge-versions-licence-id-constraints-up.sql
```diff
@@ -0,0 +1,13 @@
+/* make sure all licence_ids are populated */
+UPDATE water.charge_versions cv
+SET licence_id = l.licence_id FROM water.licences l where l.licence_ref = cv.licence_ref;
+
```
james-siteclick, line 4: If we're about to set a not null constraint here it may be safest to first delete any rows with nulls in this column

james-siteclick (changes requested):

james-siteclick (approved):

PR #1231, opened 2020-11-17
Adds is_enabled_for_new_charge_versions flag to water.change_reasons

james-siteclick:
Relating to `WATER-2874`

* Migration to add a new DB field water.change_reasons.is_enabled_for_new_charge_versions
* Adds field to model and mapper
* Defines a model constructor to set the default value to true for the flag


src/lib/models/change-reason.js
```diff
@@ -35,6 +35,15 @@ class ChangeReason extends Model {
     validators.assertEnum(type, changeReasonTypes);
     this._type = type;
   }
+
+  get isEnabledForNewChargeVersions () {
+    return this._isEnabledForNewChargeVersions;
+  }
+
+  set isEnabledForNewChargeVersions (value) {
+    validators.assertIsBoolean(value);
```
Stephan-Hough, line 44: Do you think we set a default value of true in the constructor then we don't have to worry about setting this value for new charge versions. I assume it will be very unlikely that we would want to create a new charge version that is disabled.

Stephan-Hough (approved):

PR #1232, opened 2020-11-18
Feat/water 2523 update customer details in the charge module

timtamimi:
Original ticket: https://eaflood.atlassian.net/browse/WATER-2523


src/modules/invoice-accounts/jobs/update-customer.js
```diff
@@ -0,0 +1,28 @@
...
+      retryBackoff: true
+    }
+  };
+};
+
+const handleUpdateCustomer = async job => {
+  const invoiceAccountId = get(job, 'data.invoiceAccountId');
+  return updateCustomer(invoiceAccountId);
+};
+
```
james-siteclick, line 25: Doesn't look like we have an on-complete/failed handler for this.  Just wondering what we should do if even after x attempts, the CM still does not respond.

We should probably think about the best option for handling this.

james-siteclick (changes requested):

src/modules/invoice-accounts/jobs/update-customer.js
timtamimi: Yeah, I agree - We don't currently have a defined approach for handling that case. 

For now, I'll add on-complete and failed handlers which will log an event to the console. 

More than happy to add additional functionality, but I'm not sure what that functionality would be. I thought about adding an email notification, similar to what we get from FME. E.g. "This particular address could not be updated in the charge module" - but I didn't go ahead because I don't know if that's actually helpful/desirable.

src/lib/mappers/charge-module.js
```diff
@@ -0,0 +1,81 @@
...
+      , { length: 60 }),
+    postcode: truncate(address.postcode, { length: 60 })
+  };
+};
+
+const extractCustomerName = (company = {}, agentCompany = {}) => {
+  return agentCompany.name ? agentCompany.name : company.name;
+};
+
+const extractFAO = invoiceAccount => {
```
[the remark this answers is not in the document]
timtamimi: Yup! This is much better. Didn't know this was there. Thank you.

src/modules/invoice-accounts/jobs/update-customer.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+const { get } = require('lodash');
+const { updateCustomer } = require('../../../lib/connectors/charge-module/customers-with-retry');
+const JOB_NAME = 'customer.updateAccount.*';
+
+const createMessage = async (invoiceAccountId) => {
```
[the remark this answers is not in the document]
timtamimi: Swapped out PGBoss for BullMQ. Thanks for your help on this. 🙏🏽

src/lib/mappers/charge-module.js
```diff
@@ -0,0 +1,81 @@
...
+   *   addressLine2: stringValidator.max(240).allow('', null),
+   *   addressLine3: stringValidator.max(240).allow('', null),
+   *   addressLine4: stringValidator.max(240).allow('', null),
+   *   addressLine5: stringValidator.max(60).allow('', null),
+   *   addressLine6: stringValidator.max(60).allow('', null),
+   *   postcode: stringValidator.max(60).allow('', null)
+   */
+
+  return {
+    addressLine1: truncate(`${fao ? 'FAO ' + fao + ', ' + address.addressLine1 : address.addressLine1}`, { length: 360 }),
```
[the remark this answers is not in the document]
timtamimi: I've moved the address snipper to ../lib/helpers and it is now shared between the notify functionality and the CM functionality.

src/lib/mappers/charge-module.js
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+/**
+ * Maps the invoice account model to what the charge module expects
+ * for a customer record
+ * @param {Object} invoiceAccount modelled object
+ * @return {Object} mapped customer object
+ */
+const mapInvoiceAccountToChargeModuleCustomer = async invoiceAccount => {
+  const mappedInvoiceAccount = await invoiceAccountMapper.crmToModel(invoiceAccount);
```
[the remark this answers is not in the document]
timtamimi: Nope. Just habit. I've removed the async. :)

src/lib/connectors/charge-module/customers.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+const request = require('./request');
+const invoiceAccountsConnector = require('../../../lib/connectors/crm-v2/invoice-accounts');
+const chargeModuleMappers = require('../../../lib/mappers/charge-module');
+
+const getInvoiceAccountData = async (invoiceAccountId) => {
+  const data = await invoiceAccountsConnector.getInvoiceAccountById(invoiceAccountId);
```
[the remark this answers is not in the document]
timtamimi: Good point. I've removed the fetching of the data back out to the service layer.

src/lib/connectors/charge-module/customers-with-retry.js
```diff
@@ -0,0 +1,31 @@
...
+
+const customersApi = require('./customers');
+
+/**
+ * Wraps a charge module API method so that it retries a number
+ * of times, awaiting a status other than 'generating_summary'
+ * @param {String} method - charge module bill run connector method
+ * @param  {...any} args - additional arguments
+ */
+const retry = (method, ...args) => {
```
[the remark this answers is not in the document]
timtamimi: Yup. Good point. This was done originally to keep inline with some of the other jobs I saw dotted around. This file is now removed.

src/modules/billing/jobs/update-customer.js
```diff
@@ -0,0 +1,45 @@
...
+    retryDelay: 30,
+    retryBackoff: true
+  }
+]);
+
+const createMessage = partial(messageInitialiser, JOB_NAME);
+
+const handler = async job => {
+  const invoiceAccountId = get(job, 'data.invoiceAccountId');
+  const invoiceAccountData = await invoiceAccountsConnector.getInvoiceAccountById(invoiceAccountId);
```
DGajewska, line 27: I think you should use the invoice accounts service here (`src/lib/services/invoice-accounts-service.js`). It will return the data already mapped to the model, so you won't have to map it to the model in the `mapInvoiceAccountToChargeModuleCustomer`

src/lib/mappers/charge-module.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+const { truncate, identity, pick } = require('lodash');
+const invoiceAccountMapper = require('../../lib/mappers/invoice-account');
+const { combineAddressLines, getAddressObjectFromArray } = require('./lib/helpers');
+
+/**
+ * @module maps service models to charge module expected schema
+ */
```
DGajewska, line 8: I think that these functions should live in the invoice accounts mappers file (`src/lib/mappers/invoice-account.js`)

src/lib/services/invoice-accounts-service.js
```diff
@@ -153,12 +152,13 @@ const persist = async (regionId, startDate, invoiceAccount) => {
...
 
     await createInvoiceAccount(regionId, invoiceAccount.company.id, formattedStartDate, invoiceAccount, newModels);
-
     const invoiceAccountAddress = await invoiceAccountAddressesService.createInvoiceAccountAddress(invoiceAccount, invoiceAccount.invoiceAccountAddresses[0], formattedStartDate);
+
     return invoiceAccount.fromHash({
       invoiceAccountAddresses: [invoiceAccountAddress]
     });
   } catch (err) {
+    console.log(err);
```
DGajewska, line 161: Did you mean to leave this in?

test/lib/connectors/charge-module/customers.js
```diff
@@ -0,0 +1,49 @@
...
+experiment('lib/connectors/charge-module/customers', async () => {
+  const tempInvoiceAccountId = await uuid();
+
+  beforeEach(async () => {
+    sandbox.stub(request, 'get').resolves();
+    sandbox.stub(request, 'post').resolves();
+    sandbox.stub(request, 'patch').resolves();
+    sandbox.stub(request, 'delete').resolves();
+    sandbox.stub(invoiceAccountsConnector, 'getInvoiceAccountById').resolves();
+    sandbox.stub(chargeModuleMappers, 'mapInvoiceAccountToChargeModuleCustomer').resolves(tempInvoiceAccountId);
```
DGajewska, line 29: I think you've got some redundany stubs here

DGajewska (changes requested):

src/lib/mappers/charge-module.js
timtamimi: I'd like to keep this in its own file because that's what we did for Notify's mappers. Would that be OK?

src/lib/services/invoice-accounts-service.js
timtamimi: Yes - But I'll change it to logger() as per our chat :)

src/modules/billing/jobs/update-customer.js
timtamimi: That's a great suggestion. I'll implement it. Thanks :)

test/lib/connectors/charge-module/customers.js
timtamimi: Yep you're right. Apologies. I'll clean this Test up.

src/lib/mappers/charge-module.js
DGajewska: It's hard to know what the correct convention is over such a large and sprawling code base. We do have `modelToChargeModule` mappers for `Batch` and `Transaction` models, so I saw this as the equivalent to that for the `InvoiceAccount` and `Address` models as that is where I would look for them

DGajewska (approved):

PR #1233, opened 2020-11-19
Feat/bull mq test

james-siteclick:
* Replaces job queue PG boss implementation with Bull MQ
* Removes deprecated BillingTransactions repo
* Uses job retry to manage getting charge module summary
* Removes redundant charge module API connector code
* Marks batch as errored after max retries
* Removes event ID from job queue
* Merges handler and on-complete into single file for each job


timtamimi:
### Important note
James: "We will need to ensure Redis is upgraded on test/preprod before merging"

Stephan-Hough (approved):

PR #1234, opened 2020-11-23
Uses single knex connection/pool for all DB queries

james-siteclick:
`WATER-2951`

* Uses a single `knex` instance with connection pool for all DB queries
* Adds a compatibility layer so that pool.query() can still be used for older code which does not use Bookshelf/knex and supplies an array of query params

### DIFF
--- a/src/lib/connectors/db.js
+++ b/src/lib/connectors/db.js
@@ -1,16 +1,31 @@
-require('dotenv').config();
-const pg = require('pg');
-const moment = require('moment');
-const helpers = require('@envage/water-abstraction-helpers');
+'use strict';
 
-const config = require('../../../config.js');
-const { logger } = require('../../logger');
+const knex = require('./knex');
 
-const DATE_FORMAT = 'YYYY-MM-DD';
-const dateMapper = str => moment(str).format(DATE_FORMAT);
+/**
+ * Allows a query to be run with same arguments
+ * as pg.pool.query(query, params)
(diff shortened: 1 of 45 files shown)


src/lib/connectors/db.js
```diff
@@ -1,16 +1,32 @@
...
+const query = (sqlQuery, params = []) => {
+  if (params.length === 0) {
+    return knex.knex.raw(sqlQuery);
+  }
 
-// Set dates to format YYYY-MM-DD rather than full date/time string with timezone
-pg.types.setTypeParser(pg.types.builtins.DATE, dateMapper);
+  const data = params.reduce((acc, param, i) => {
+    acc.params[`param_${i}`] = param;
+    const r = new RegExp(`\\$${(i + 1)}(?![0-9])`, 'g');
```
timtamimi, line 22: I'm not confident I understand what this regex is doing. May I recommend adding an inline comment with a synopsis of its purpose please? I assume it's in some way parsing the query syntax so that it can be passed through knex.raw, but I'm not sure what the transformation is.

timtamimi (approved):
Looks generally good and I can see what it's doing at a high level, so approved 👍🏽  , but please consider adding a description around the Regex to explain what it's doing as I was less certain about what's happening there.

src/lib/connectors/db.js
james-siteclick: Thanks I have added comments and tweaked the code slightly

PR #1235, opened 2020-11-25
Feat: Updates minimum charge transactions with charge data 

DGajewska:
WATER-2850

* Creates transaction CM connector
* Calls CM for additional data for each minimum charge transaction
* Updates cmToModel transaction mapper to include new details

### DIFF
--- a/src/modules/billing/mappers/charge-module-decorators.js
+++ b/src/modules/billing/mappers/charge-module-decorators.js
@@ -59,14 +59,21 @@ const decorateTransactions = async (invoiceLicence, cmTransactions) => {
   if (isEmpty(invoiceLicence.transactions)) return invoiceLicence;
   for (const cmTransaction of cmTransactions) {
     const serviceTransaction = invoiceLicence.transactions.find(trans => trans.externalId === cmTransaction.id);
+
+    // CM transactions should have corresponding transactions in the service,
+    // decorate these with data from CM
     if (serviceTransaction) {
       decorateTransaction(serviceTransaction, cmTransaction);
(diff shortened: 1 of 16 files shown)


src/modules/billing/mappers/charge-module-decorators.js
```diff
@@ -44,25 +45,30 @@ const decorateTransaction = (serviceTransaction, cmTransaction) =>
...
+  for (const cmTransaction of cmTransactions) {
     const serviceTransaction = invoiceLicence.transactions.find(trans => trans.externalId === cmTransaction.id);
     if (serviceTransaction) {
-      return decorateTransaction(serviceTransaction, cmTransaction);
-    }
-    if (!cmTransaction.minimumChargeAdjustment) {
+      decorateTransaction(serviceTransaction, cmTransaction);
+    } else if (!cmTransaction.minimumChargeAdjustment) {
+      // the only transactions which do not exist in the service should be minimum charge
       throw new TransactionStatusError(`Unexpected Charge Module transaction externalId: ${cmTransaction.id}`);
```
timtamimi, line 66: Thanks for walking me through this. Shall we just move the error throwing to the end of the loop? It's not mandatory - Just good convention I think.

timtamimi (changes requested):
One change suggested.

src/modules/billing/mappers/charge-module-decorators.js
DGajewska: Yes, I think that it would make it a tad bit clearer as well :)

timtamimi (approved):

PR #1242, opened 2020-12-07
Feat: Proxy creating address in CRM

DGajewska:
WATER-2953

* Adds address module
* Calls CRM to create address
* Returns address mapped to model
* Updates address mapper


src/modules/addresses/routes.js
```diff
@@ -0,0 +1,35 @@
...
+          address1: OPTIONAL_STRING,
+          address2: OPTIONAL_STRING,
+          address3: OPTIONAL_STRING,
+          address4: OPTIONAL_STRING,
+          town: OPTIONAL_STRING,
+          county: OPTIONAL_STRING,
+          country: Joi.string().required(),
+          postcode: OPTIONAL_STRING,
+          isTest: Joi.boolean().optional().default(false),
+          dataSource: Joi.string().valid('wrls', 'nald').default('wrls'),
```
Stephan-Hough, line 26: need to add companiesHouse as a source

Stephan-Hough (changes requested):
companiesHouse should be added as a dataSource to the route

Stephan-Hough (approved):

PR #1257, opened 2020-12-14
Send unique transaction ID to CM - WATER 2945

Stephan-Hough:
feat/water-2945
-- create charge job modified to;
-- retry if job fails
-- if the error is 409 update the transaction with the external id
-- if retry count exceeded then set transaction to error and continue


src/modules/billing/jobs/create-charge.js
```diff
@@ -26,7 +28,12 @@ const createMessage = (batchId, billingBatchTransactionId) => ([
     billingBatchTransactionId
   },
   {
-    jobId: `${JOB_NAME}.${batchId}.${billingBatchTransactionId}`
+    jobId: `${JOB_NAME}.${batchId}.${billingBatchTransactionId}`,
+    attempts: 10,
```
james-siteclick, line 32: I think maybe 6 attempts might be enough here - that's about 5 minutes

src/modules/billing/jobs/create-charge.js
```diff
@@ -82,20 +88,25 @@ const handler = async job => {
...
+    // if error code >= 400 and < 500 set transacti on status to error and continue
+    if (isClientError(err)) {
+      batchJob.logHandlingError(job, err);
+      transactionsService.setErrorStatus(transactionId);
+      return {
+        batch: job.data.batch
+      };
     }
+    // throw error to retry
+    throw new StateError(`Charge not created in CM for transaction ${transactionId} in batch ${batchId}`);
```
james-siteclick, line 108: I think we should possibly rethrow `err` here instead of creating a new error.  Otherwise e.g. if it's a DB error our side this will be masked.

src/modules/billing/jobs/create-charge.js
```diff
@@ -82,20 +88,25 @@ const handler = async job => {
...
-
-    // If not a client error, error the batch
-    if (!isClientError(err)) {
-      await batchService.setErrorStatus(batchId, BATCH_ERROR_CODE.failedToCreateCharge);
-      throw err;
+    const statusCode = get(err, 'statusCode');
+
+    // if the charge was created in the CM
+    if (statusCode === 409) {
+      batchJob.logHandlingError(job, err);
```
james-siteclick, line 95: May as well move:
```
 batchJob.logHandlingError(job, err);
```
to the top level catch block, then we don't need to repeat it below

src/modules/billing/jobs/create-charge.js
```diff
@@ -113,9 +124,31 @@ const onComplete = async (job, queueManager) => {
...
+  } else if (err.name === 'StateError') {
+    // Only logger an info message if we the CM has not created a charge - this is expected
+    // behaviour
+    logger.info(err.message);
+  } else {
+    // Do normal error logging
+    helpers.onFailedHandler(job, err);
+  }
+};
+
```
james-siteclick, line 148: I reckon we could maybe just do:

```
const onFailedHandler = async (job, err) => {
  const batchId = get(job, 'data.batchId');
  const transactionId = get(job, 'data.billingBatchTransactionId');

  // On final attempt, error the batch and log
  if (helpers.isFinalAttempt(job)) {
    try {
      logger.error(`Transaction with id ${transactionId} not generated in CM after ${job.attemptsMade} attempts, marking batch as errored ${batchId}`);
      await batchService.setErrorStatus(batchId, BATCH_ERROR_CODE.failedToPrepareTransactions);
    } catch (error) {
      logger.error(`Unable to set batch status ${batchId}`, error);
    }
  } 

  return helpers.onFailedHandler(job, err);
};

```

james-siteclick (changes requested):

james-siteclick (approved):

PR #1261, opened 2020-12-16
Fix: Updates contact schema validation

DGajewska:
WATER-2963


src/lib/mappers/contact.js
```diff
@@ -39,8 +39,8 @@ const uiToModel = contactData => {
     return new Contact(contactData.contactId);
   }
   const contact = new Contact();
-  contact.dataSource = Contact.DATA_SOURCE_TYPES.wrls;
-  return contact.fromHash(contactData);
+  contact.dataSource = contactData.source || Contact.DATA_SOURCE_TYPES.wrls;
```
Stephan-Hough, line 42: It would be better if default values are applied in the model constructor then you don't need this line, the route validation also sets the default value to wrls so this is unlikely to be undefined/null?

Stephan-Hough (approved):
Approved with one optional minor observation for you to consider.

src/lib/mappers/contact.js
DGajewska: Good catch! I'll have a look at how we apply this elsewhere in the service

DGajewska: It appears that this mapper is used in a couple of different places, the source is set in the validation in one place and not in the other, so I'm going to leave this as is

PR #1269, opened 2021-01-18
Feat: Bulk Upload Y/N fields accept Yes/No

DGajewska:
WATER-2794


src/modules/returns/lib/csv-adapter/validator.js
```diff
@@ -169,7 +169,7 @@ const validateReturnReference = licence => {
 
 const isValidBooleanOrEmpty = (licence, licenceKey, errorKey = licenceKey) => {
   const { value, line } = licence[licenceKey];
-  const isValid = ['y', 'n', ''].includes(value.toLowerCase());
+  const isValid = ['y', 'n', 'yes', 'no', ''].includes(value.toLowerCase());
```
timtamimi, line 172: Is it worth changing `value.toLowerCase()` to `value.toLowerCase().trim()`?

DGajewska: Oooh, yes. Thanks for that. Updated :)

PR #1273, opened 2021-01-21
Feat/water 2903 bills tab

timtamimi:
Original ticket: https://eaflood.atlassian.net/browse/WATER-2903

- Adds endpoint for grabbing invoices associated to a licence
- Handles pagination using an existing helper


src/modules/licences/controllers/invoices.js
```diff
@@ -0,0 +1,17 @@
...
+const { logger } = require('../../../logger');
+
+const getLicenceInvoices = async request => {
+  const { licenceId } = request.params;
+  const { page, perPage } = request.query;
+  try {
+    return licencesService.getLicenceInvoices(licenceId, page, perPage);
+  } catch (error) {
+    logger.error('Failed to get bills for licence', error, { licenceId });
+    return Boom.boomify(error);
```
DGajewska, line 13: I don't think that this reads the error status code, it must be explicitly provided, otherwise Boom will default to a 500, which could cause some confusion. eg. if the error status code is actually 404, the result would be:
```
Error: message
... {
statusCode: 404,
options: {
    statusCode: 500,
    ...
}}
```

We have an error response mapper which is used in `src/modules/licences/controllers/agreements.js` which might make things easier :)

DGajewska (changes requested):

DGajewska (approved):

PR #1274, opened 2021-01-22
Bugfix: Marking billing volumes are errored not required.

DGajewska:
WATER-2991

Updates call to billing volumes service when marking batch as errored making the update not required as this throws an error when no rows are updated.

### DIFF
--- a/src/lib/connectors/repos/billing-volumes.js
+++ b/src/lib/connectors/repos/billing-volumes.js
@@ -142,12 +142,12 @@ const updateByBatchId = async (billingBatchId, changes) => {
  * Marks all records as errored by batch ID
  * @param {String} billingBatchId
  */
-const markVolumesAsErrored = async (billingBatchId) => {
+const markVolumesAsErrored = async (billingBatchId, options = {}) => {
   const result = await BillingVolume
     .where('billing_batch_id', billingBatchId)
     .save({
       errored_on: new Date()
-    }, { method: 'update' });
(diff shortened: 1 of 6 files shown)


src/lib/connectors/repos/billing-volumes.js
```diff
@@ -142,12 +142,12 @@ const updateByBatchId = async (billingBatchId, changes) => {
  * Marks all records as errored by batch ID
  * @param {String} billingBatchId
  */
-const markVolumesAsErrored = async (billingBatchId) => {
+const markVolumesAsErrored = async (billingBatchId, options = {}) => {
```
timtamimi, line 145: Do you think it's worth changing the default for `options` from `{}` to `{required: false}`?

timtamimi (approved):
I left one optional question/suggested change. Happy for this to be merged either ways.

src/lib/connectors/repos/billing-volumes.js
DGajewska: I don't know if we would want to make it the default behaviour. Also, I think that the function call `markVolumesAsErrored(batchId, { require: false })` is clearer than `markVolumesAsErrored(batchId, {})` when providing the second variable. 

I have done this same type of thing with a `isUpdateRequired` boolean flag as the second parameter which defaults to false, but I thought that `options = {}` was most flexible and provided the best readability.

PR #1277, opened 2021-01-27
Tweak billing Bull MQ jobs to return less data and remove on complete…

james-siteclick:
WATER-2997
* Modifies Bull job configuration to remove jobs on complete/failure
* Modifies billing jobs to only require/return a minimally needed set of data for storage in Redis for efficiency

### DIFF
--- a/src/lib/message-queue-v2/QueueManager.js
+++ b/src/lib/message-queue-v2/QueueManager.js
@@ -5,6 +5,11 @@ const bull = require('bullmq');
 const STATUS_COMPLETED = 'completed';
 const STATUS_FAILED = 'failed';
 
+const jobDefaults = {
+  removeOnComplete: true,
+  removeOnFail: 100
+};
+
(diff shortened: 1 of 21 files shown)


src/lib/message-queue-v2/QueueManager.js
```diff
@@ -5,6 +5,11 @@ const bull = require('bullmq');
 const STATUS_COMPLETED = 'completed';
 const STATUS_FAILED = 'failed';
 
+const jobDefaults = {
+  removeOnComplete: true,
+  removeOnFail: 100
```
timtamimi, line 10: I think 100 might be low. May I suggest 500ish?

timtamimi (changes requested):

src/lib/message-queue-v2/QueueManager.js
james-siteclick: Done

PR #1283, opened 2021-02-04
Fix: Call CM multiple times for paginated transaction results

DGajewska:
WATER-2976


test/modules/billing/services/batch-service.js
```diff
@@ -1434,4 +1435,49 @@ experiment('modules/billing/services/batch-service', () => {
...
+      chargeModuleBillRunConnector.getTransactions.onSecondCall().resolves({
+        data: getTransaction(2)
+      });
+      chargeModuleBillRunConnector.getTransactions.onThirdCall().resolves({
+        data: getTransaction(3)
+      });
+
+      result = await batchService.getAllCmTransactionsForBatch({ externalId: 'test-cm-batch-id' });
+    });
+
```
james-siteclick, line 1460: Might be good to have a little test here to check that:
```
chargeModuleBillRunConnector.getTransactions.callCount
```
has been called the expected number of times based on the number of pages expected

james-siteclick (approved):

test/modules/billing/services/batch-service.js
DGajewska: Thank you :) I wanted to do a test like this, but must've got distracted and forgot about it in the end! I've added it in now

PR #1298, opened 2021-02-24
Fix supplementary billing to cope with imported NALD transactions

james-siteclick:
`WATER-3030`

Creates a new supplementary billing process which can deal with either historical NALD transactions or WRLS transactions.

* Historical and current batch transactions are compared on billable days/volumes within a group of similar transactions rather than looking for identical transactions
* Grouping keys are calculated on-the-fly removing the need for the `transaction_key` column
* Detects pairs of historical transactions that already cancel each other and does not reverse the pair
* Can reverse historical charges and credits rather than just charges


src/modules/billing/services/supplementary-billing-service/data-service.js
```diff
@@ -0,0 +1,138 @@
...
+
+const { actions } = require('./constants');
+
+/**
+ * These are the keys present on a transaction object
+ * from the initial DB query and subsequent processing
+ * which must be omitted before attempting to write the
+ * transaction back to the DB
+ */
+const omitKeys = [
```
Stephan-Hough, line 21: Would it not be safer to just pick the data you need. What if later on someone adds another key to objects this might then fail?

src/modules/billing/services/supplementary-billing-service/supplementary-processor.js
```diff
@@ -0,0 +1,302 @@
...
+/**
+ * Gets a key for pairing transactions within a group which have the same:
+ * - Charge period
+ * - Billable days (for annual transactions)
+ * - Volume (for TPT transactions)
+ *
+ * @param {Object} transaction
+ * @return {String}
+ */
+const getPairGroupingKey = transaction => {
```
Stephan-Hough, line 176: when the transactions are grouped and filtered - do they not need to be grouped by charge element Id or invoice licence Id?

Stephan-Hough (changes requested):
A couple of questions posted - I might be missing something but grateful if we could catchup on this PR thanks

src/modules/billing/services/supplementary-billing-service/supplementary-processor.js
james-siteclick: No, we are looking for:
* Transactions that have come from a mostly identical charge element, but not with the same ID
* The historical transactions won't be on the same invoice or invoice licence as the current batch ones as they are spread across multiple billing batches

src/modules/billing/services/supplementary-billing-service/data-service.js
james-siteclick: Agreed, I'll change this

Stephan-Hough (approved):

PR #1302, opened 2021-03-04
Feat: Delete charge info workflows

DGajewska:
WATER-3002

* Removes redudant licence methods
* Tweaks charge version workflow methods to work with new model set up
* Adds test data to enable regression tests on charge version workflows


src/modules/charge-versions/controllers/charge-versions.js
```diff
@@ -62,13 +62,13 @@ const postCreateFromWorkflow = async request => {
   const { chargeVersionWorkflowId } = request.params;
   try {
     const chargeVersionWorkflow = await chargeVersionWorkflowService.getById(chargeVersionWorkflowId);
-
+    console.log({ chargeVersionWorkflow });
```
Stephan-Hough, line 65: We don't need this line

src/modules/charge-versions/controllers/charge-versions.js
```diff
@@ -62,13 +62,13 @@ const postCreateFromWorkflow = async request => {
...
     if (!chargeVersionWorkflow) {
       return Boom.notFound(`Charge version workflow ${chargeVersionWorkflowId} not found`);
     }
 
     const approvedBy = userMapper.pojoToModel(request.defra.internalCallingUser);
-
+    console.log({ approvedBy });
```
Stephan-Hough, line 71: and this one

Stephan-Hough (changes requested):
There are a couple of console logs that can be removed from src/modules/charge-versions/controllers/charge-versions.js

src/modules/charge-versions/controllers/charge-versions.js
DGajewska: Thank you :) removed

src/modules/charge-versions/controllers/charge-versions.js
DGajewska: Thank you :) removed

Stephan-Hough (approved):

PR #1313, opened 2021-03-09
Fix: Makes approving billing volumes not required

DGajewska:
WATER-3069

* If all billing volumes are removed in Review stage, will not throw error

### DIFF
--- a/src/lib/connectors/repos/billing-volumes.js
+++ b/src/lib/connectors/repos/billing-volumes.js
@@ -131,10 +131,10 @@ const deleteByBatchIdAndLicenceId = (billingBatchId, licenceId) =>
  * @param {String} billingBatchId
  * @param {Object} changes
  */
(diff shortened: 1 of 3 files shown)


src/lib/connectors/repos/billing-volumes.js
```diff
@@ -131,10 +131,10 @@ const deleteByBatchIdAndLicenceId = (billingBatchId, licenceId) =>
  * @param {String} billingBatchId
  * @param {Object} changes
  */
-const updateByBatchId = async (billingBatchId, changes) => {
+const updateByBatchId = async (billingBatchId, changes, options = {}) => {
```
james-siteclick, line 134: As this function is only called in 1 place, we could only add `{ require: false }` to the call below rather than exposing Bookshelf options to the invoking code

james-siteclick (approved):

PR #1314, opened 2021-03-09
Feat: Adds non-chargeable reason for licence revoked a month after annual billing

DGajewska:
WATER-3040

* Adds new non-chargeable change reason
* Removes position column in water.change_reasons


test/lib/connectors/repos/change-reasons.js
```diff
@@ -43,7 +43,7 @@ experiment('lib/connectors/repos/change-reasons', () => {
     });
 
     test('collection is sorted on position property', async () => {
```
james-siteclick, line 45: We may need to tweak the test description to 
`sorted on description` 
to match the test

james-siteclick (changes requested):

test/lib/connectors/repos/change-reasons.js
DGajewska: Oh yes! Thank you :)

james-siteclick (approved):

PR #1329, opened 2021-03-19
Fix: Updates Bills tab

DGajewska:
WATER-3042

* Adds PSC user to test data
* Makes acceptance test services consistent. Uses bookshelf
* Creates billing test data
* Adds services to delete test billing data
* Adds missing DB fields to invoice model and mapper
* Updates query and mapping to get all invoices for a licence


src/lib/connectors/repos/billing-invoice-licences.js
```diff
@@ -34,18 +34,8 @@ const findOne = async id => {
...
-    })
-    .orderBy('date_created', 'DESC')
-    .fetchPage({
-      pageSize: perPage,
-      page: page,
-      withRelated
-    });
-  return paginationHelper.paginatedEnvelope(result);
+  const results = await raw.multiRow(queries.findAllByLicenceIdForSentBatches, { licenceId });
+  return paginationHelper.paginateRawQueryResults(results, page, perPage);
```
james-siteclick, line 38: We should ideally not be fetching all rows from the DB if possible, and relying on the DB to get a page.

The query below looks fairly simple - I think it might be possible to create the query with Bookshelf and use Bookshelf's pagination features.

src/lib/services/licences.js
```diff
@@ -131,6 +131,30 @@ const getLicenceAccountsByRefAndDate = async (documentRef, date) => {
 const flagForSupplementaryBilling = licenceId =>
   repos.licences.update(licenceId, { includeInSupplementaryBilling: INCLUDE_IN_SUPPLEMENTARY_BILLING.yes });
 
+const getLabelForInvoiceWithoutInvoiceNumber = invoice => {
```
james-siteclick, line 134: I feel like this logic might live as a getter on the model itself, e.g.
```
get displayLabel() {
  if(this.invoiceNumber) {
   return this.invoiceNumber;
  }
  // logic below...
}
```

It's necessary to override the `toJSON()` method in this case to include the above calculated property - this is done on a few of the other models.

james-siteclick (changes requested):

src/lib/services/licences.js
DGajewska: I thought this as well and tried it out, but I must've done something wrong because it wasn't working for me, but I will give it another go

james-siteclick (approved):

PR #1335, opened 2021-03-25
Feat: Stores isFactorsOverridden flag when creating new charge element

DGajewska:
WATER-3072

* Calculates value of flag
* Updates model and mapper


src/lib/mappers/charge-element.js
```diff
@@ -26,6 +26,7 @@ const dbToModelMapper = createMapper()
     'chargeVersionId'
   )
   .map().to('abstractionPeriod', abstractionPeriodMapper.dbToModel)
+  .map('factorsOverridden').to('isFactorsOverridden', factorsOverridden => factorsOverridden)
```
james-siteclick, line 29: I think the second argument to `.to()` can be omitted if you don't need to map the value

james-siteclick (changes requested):

src/lib/mappers/charge-element.js
DGajewska: Ah yes! I did have something going on in the function, but took it out. That looks quite silly 😂

james-siteclick (approved):

PR #1339, opened 2021-03-26
Water-3070 Charge version workflow tweaks

Stephan-Hough:
**Feat/water 3070 cv wf tweaks**

- pre handler removed from post Charge Version workflow - licence version is now set to null
- convert to cubic meters now returns 0 instead of null to avoid charge version workflow breaking when authorised quantity is null on abstraction data


src/modules/charge-versions/controllers/charge-version-workflow.js
```diff
@@ -66,14 +59,21 @@ const postChargeVersionWorkflow = async request => {
...
   const { chargeVersion } = request.pre;
 
   const changes = {
     ...request.payload,
     ...chargeVersion && { chargeVersion }
   };
-
+  if (createdBy) {
+    try {
+      changes.createdBy = userMapper.pojoToModel(createdBy);
```
DGajewska, line 71: Does this need a try-catch? If we use the calling user found in `request.defra.internalCallingUser`, that data is taken from the IDM, so we can be sure that the data provided is in the correct format

This is the approach that is taken when creating charge versions from workflows in `controllers/charge-versions.js`

src/lib/unit-conversion.js
```diff
@@ -4,7 +4,7 @@ const { isNull } = require('lodash');
 
 const cubicMetresToMegalitres = value => {
   if (isNull(value)) {
-    return null;
+    return 0;
```
DGajewska, line 7: I'm not sure that we can do this because in certain cases, null and 0 mean two different things

src/modules/charge-versions/routes/charge-version-workflows.js
```diff
@@ -90,7 +89,8 @@ module.exports = {
         payload: {
           status: Joi.string(),
           chargeVersion: Joi.object(),
-          approverComments: Joi.string().allow(null)
+          approverComments: Joi.string().allow(null),
+          createdBy: Joi.object().optional()
```
DGajewska, line 93: Would not require this if we use the internal calling user

src/modules/charge-versions/routes/charge-version-workflows.js
```diff
@@ -90,7 +89,8 @@ module.exports = {
         payload: {
```
DGajewska, line 89: Would be good to add `Joi.object()` while you're here

DGajewska (changes requested):

src/modules/charge-versions/controllers/charge-version-workflow.js
Stephan-Hough: The reason why we can't use request.defra object is because this endpoint can be used to update the charge version workflow record at anytime of the process so it will overwrite the original creator. I use the defra.user object in the UI though to get this info for the first submission of the draft charge info

src/modules/charge-versions/routes/charge-version-workflows.js
Stephan-Hough: as mentioned above - unfortunately it is the only option with the current workflow/cron job solution

DGajewska (approved):

PR #1347, opened 2021-04-06
Feat: Endpoint for flagging invoices for rebilling

DGajewska:
WATER-3053

* Reorganises billing route and controller files
* Adds migrations to add rebilling flag to water.billing_invoices
* Adds isFlaggedForRebilling flag to invoice model and mapper
* Adds route and controller for PATCH endpoint
* Adds service function 
* Tweaks repo function


src/modules/billing/routes/invoices.js
```diff
@@ -0,0 +1,25 @@
...
+  method: 'PATCH',
+  path: '/water/1.0/billing/invoices/{invoiceId}',
+  handler: controller.patchInvoice,
+  config: {
+    validate: {
+      params: Joi.object({
+        invoiceId: Joi.string().guid().required()
+      }),
+      payload: Joi.object({
+        isFlaggedForRebilling: Joi.boolean().optional()
```
Stephan-Hough, line 18: if isFlaggedForRebilling is the only property you can update then it should be required

DGajewska: I did think that as well, but everything else is done in a much more generic way, but maybe that's because I'm reusing a lot of helper methods. I'm happy to change this :) Thank you

Stephan-Hough (approved):

PR #1353, opened 2021-04-08
Feat/water 2864

timtamimi:
Service PR: https://github.com/DEFRA/water-abstraction-service/pull/1353
UI PR: https://github.com/DEFRA/water-abstraction-ui/pull/1672
Reporting PR: https://github.com/DEFRA/water-abstraction-reporting/pull/4

Main ticket: https://eaflood.atlassian.net/browse/WATER-2864


src/lib/services/s3.js
```diff
@@ -61,7 +61,22 @@ const download = async (key, destination) => {
  */
 const getHead = key => callS3MethodWithKey('headObject', key);
 
+/**
+ * Gets a signed URL for a particular file
+ * @param {String} filepath
+ * @param {String} operation
+ */
+const getSignedUrl = (filepath, operation = 'getObject', expiry = 60) => {
```
james-siteclick, line 69: Not sure this is used any more?

src/modules/reporting/controllers.js
```diff
@@ -0,0 +1,5 @@
+const reportsConnector = require('../../lib/connectors/reporting/');
+
+const getReport = request => reportsConnector.getReport(request.params.reportIdentifier);
```
james-siteclick, line 3: As with UI layer, I think it would be more efficient to pipe this if the reporting module is responding with a CSV file

james-siteclick (changes requested):

src/lib/services/s3.js
timtamimi: You're right. I was keeping it on the off chance it is useful in the future but I'll take it out

src/lib/connectors/reporting/index.js
```diff
@@ -0,0 +1,24 @@
...
+const config = require('../../../../config');
+const urlJoin = require('url-join');
+const got = require('got');
+
+/**
+ * Requests a report from the Reporting microservice
+ * @param request {Object} The request object
+ * @returns {Request} Got Request - Streams file from S3
+ */
+const getReport = request => {
```
james-siteclick, line 11: As with the UI one, I think this function should just take the needed arguments something like
```
getReport(key, userId) {
...
```

james-siteclick (changes requested):

src/lib/connectors/reporting/index.js
timtamimi: Done in 3ee7c6cb

james-siteclick (approved):

PR #1360, opened 2021-04-21
Feat/water 3090 rebilling

james-siteclick:
`WATER-3090`

* Adds migration to add new fields needed for re-billing processing
* Adds re-billing Bull MQ job for supplementary runs only to process re-billed invoices
* Resets re-billing flags when bill run is sent
* Includes linked bills in API response

### DIFF
--- a/src/lib/connectors/bookshelf/BillingInvoice.js
+++ b/src/lib/connectors/bookshelf/BillingInvoice.js
@@ -15,5 +15,9 @@ module.exports = bookshelf.model('BillingInvoice', {
 
   billingInvoiceLicences () {
     return this.hasMany('BillingInvoiceLicence', 'billing_invoice_id', 'billing_invoice_id');
+  },
+
+  linkedBillingInvoices () {
+    return this.hasMany('BillingInvoice', 'original_billing_invoice_id', 'original_billing_invoice_id');
   }
 });
--- a/src/lib/connectors/io-redis.js
+++ b/src/lib/connectors/io-redis.js
@@ -4,7 +4,7 @@ const IORedis = require('ioredis');
 const config = require('../../../config');
 
 // Note: this limit will need increasing if further Bull MQ job queues are added
-const maxListenerCount = 20;
+const maxListenerCount = 23;
 
 exports.createConnection = () => {
   const ioRedis = new IORedis(config.redis);
--- a/src/lib/connectors/io-redis.js
+++ b/src/lib/connectors/io-redis.js
@@ -3,11 +3,8 @@
 const IORedis = require('ioredis');
 const config = require('../../../config');
 
-// Note: this limit will need increasing if further Bull MQ job queues are added
-const maxListenerCount = 23;
-
 exports.createConnection = () => {
-  const ioRedis = new IORedis(config.redis);
-  ioRedis.setMaxListeners(maxListenerCount);
+  const ioRedis = new IORedis(config.redis.connection);
+  ioRedis.setMaxListeners(config.redis.maxListenerCount);
   return ioRedis;
 };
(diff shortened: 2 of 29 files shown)


src/lib/connectors/bookshelf/BillingInvoice.js
```diff
@@ -15,5 +15,9 @@ module.exports = bookshelf.model('BillingInvoice', {
 
   billingInvoiceLicences () {
     return this.hasMany('BillingInvoiceLicence', 'billing_invoice_id', 'billing_invoice_id');
+  },
+
+  linkedBillingInvoices () {
+    return this.hasMany('BillingInvoice', 'original_billing_invoice_id', 'original_billing_invoice_id');
```
timtamimi, line 21: Forgive my pedantry - I think this comes down to my lack of experience with Bookshelf, but

This suggests that an invoice can have numerous `original_billing_invoice_id`s

What I expected: An invoice can have one `original_billing_invoice_id`, rather than many. 

So I think my question is: Is this intentionally `hasMany`?

src/lib/connectors/io-redis.js
```diff
@@ -4,7 +4,7 @@ const IORedis = require('ioredis');
 const config = require('../../../config');
 
 // Note: this limit will need increasing if further Bull MQ job queues are added
-const maxListenerCount = 20;
+const maxListenerCount = 23;
```
timtamimi, line 7: Wondering if this should be stored in and be fetched from `config.redis.maxListeners` because I have a feeling we will end up changing it every now and then.

src/lib/models/batch.js
```diff
@@ -33,7 +33,8 @@ const BATCH_ERROR_CODE = {
   failedToCreateBillRun: 50,
   failedToDeleteInvoice: 60,
   failedToProcessTwoPartTariff: 70,
-  failedToGetChargeModuleBillRunSummary: 80
+  failedToGetChargeModuleBillRunSummary: 80,
+  failedToProcessRebilling: 90
```
timtamimi, line 37: This isn't something I want you to change - just making an observation: Those codes don't need to be codes. In the longer term, we might want to change those such that we simply store the name of the error type. E.g. Instead of `90` we would simply store `failedToProcessRebilling`.

Just a thought for the future. Not something to worry about in the context of this PR.

src/modules/billing/controllers/invoices.js
```diff
@@ -8,13 +8,19 @@ const invoiceIsPartOfSentBatch = invoice => {
...
+  invoice.rebillingState !== null;
+
 const patchInvoice = async request => {
   try {
     const { invoiceId } = request.params;
     const invoice = await invoiceService.getInvoiceById(invoiceId);
     if (!invoiceIsPartOfSentBatch(invoice)) {
       return Boom.conflict('Cannot update invoice that is not part of a sent batch');
     }
+    if (invoiceIsRebill(invoice)) {
```
timtamimi, line 21: Is this right? Wouldn't we want to patch the invoice as part of the CM refresh routine?

timtamimi (changes requested):
Left a bunch of comments!

src/lib/connectors/bookshelf/BillingInvoice.js
james-siteclick: There is a single `original_billing_invoice_id` column, but this function here is describing the relationship.

For any one of the 3 invoices (original, reversal, rebill), this relationship points from one to all 3 which allows all the linked invoices to be retrieved with a Bookshelf query. 

Don't know if that explains it?

src/lib/models/batch.js
james-siteclick: True. 
I'd also like the UI to show the error then we wouldn't have to fish around in the DB.

src/modules/billing/controllers/invoices.js
james-siteclick: This is preventing it from being flagged, rather than something that happens during the billing process.

The UI should not present invalid invoices for flagging, but this is an additional safety check.

src/lib/connectors/io-redis.js
james-siteclick: Done

timtamimi (approved):

PR #1361, opened 2021-04-23
WATER-3041 - Migration to add new notify template

Stephan-Hough:
feat/water-3041
-- migration to add notify service alert template


Stephan-Hough:
no longer needed - code shifted to the import repo

PR #1373, opened 2021-05-05
UAT bug fix - send/approve batch WATER 3142

Stephan-Hough:
feat/water-3142
-- new job added to approve batch
-- tests added

### DIFF
--- a/src/modules/billing/services/batch-service.js
+++ b/src/modules/billing/services/batch-service.js
@@ -171,6 +171,8 @@ const approveBatch = async (batch, internalCallingUser) => {
     return batch;
   } catch (err) {
     logger.error('Failed to approve the batch', err, batch);
+    // set the status back to ready so the user can retry
+    setStatus(batch.id, BATCH_STATUS.ready);
     await saveEvent('billing-batch:approve', 'error', internalCallingUser, batch);
     throw err;
   }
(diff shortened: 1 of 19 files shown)


src/modules/billing/services/batch-service.js
```diff
@@ -168,7 +168,7 @@ const approveBatch = async (batch, internalCallingUser) => {
     await licencesService.updateIncludeInSupplementaryBillingStatusForSentBatch(batch.id);
     await invoiceService.resetIsFlaggedForRebilling(batch.id);
 
-    return setStatus(batch.id, BATCH_STATUS.processing);
+    return batch;
   } catch (err) {
     logger.error('Failed to approve the batch', err, batch);
```
james-siteclick, line 173: Do you think we need to update the batch status in this `catch()` block?

I'm wondering if we might need to set it to `error`.

james-siteclick (approved):

src/modules/billing/services/batch-service.js
Stephan-Hough: I think we touched on this point when we discussed the proposed solution and I believe we agreed to not set it to error so that the user can retry to send/approve the batch if the process failed? If we set the batch to error they would have to re-run the batch. I am not sure - but happy to add it if you think we should

Stephan-Hough: However the batch status will be 'processing' when it reaches this code. so if the process fails here we might need to set it back to ready?

PR #1374, opened 2021-05-05
Feat/water 3081 informing CM when there has been a NALD linked entities change

timtamimi:
This set of pull requests enables the service to inform the charge module when there has been an entity change in the NALD data which is relevant to an invoice account.

Service pull request: https://github.com/DEFRA/water-abstraction-service/pull/1374
CRM pull request: https://github.com/DEFRA/water-abstraction-tactical-crm/pull/780
Import pull request: https://github.com/DEFRA/water-abstraction-import/pull/299

Original ticket: https://eaflood.atlassian.net/browse/WATER-3081


Stephan-Hough (changes requested):
My only concern is if the NALD import failed then this job could run again and send unnecessary updates to the Charge Module. The update customer job could update the hashes in the CRM after the updates were successfully sent to the charge module. 

If this is not an issue then I am would be happy to approve.

Stephan-Hough (approved):

PR #1397, opened 2021-05-14
Adds route and controller + associated tests

timtamimi:
Original ticket: https://eaflood.atlassian.net/browse/WATER-3146

Adds a new route `GET /water/1.0/gauging-stations/{guid}`


src/modules/gauging-stations/controllers.js
```diff
@@ -0,0 +1,8 @@
+const gaugingStationsRepo = require('../../lib/connectors/repos/gauging-stations');
+
+const getGaugingStation = request => {
+  const { stationGuid } = request.params;
+  return gaugingStationsRepo.findOne(stationGuid);
```
james-siteclick, line 5: I wonder if we might want to add 404 handling here along the lines of:
```
const gaugingStation = await gaugingStationsRepo.findOne(stationGuid);
return gaugingStation || Boom.notFound(`Gauging station ${stationGuid} not found`);
```

james-siteclick (changes requested):

PR #1399, opened 2021-05-17
Initial work to support licence summary page by licence ID

james-siteclick:
`WATER-2880`

* Adds additional endpoints by licence ID to support licence summary page.
* Modifies existing endpoints to include Licence service models (internal search, user status)


src/lib/connectors/repos/scheduled-notifications.js
```diff
@@ -26,8 +27,36 @@ const update = (id, changes) => helpers.update(ScheduledNotification, 'id', id,
...
+/**
+ * Finds scheduled notifications by licence number
+ * Checks if messages are delivered/received in Notify
+ *
+ * @param {String} licenceNumber
+ * @param {Number} page
+ * @param {Number} perPage
+ * @returns
+ */
+const findByLicenceNumber = async (licenceNumber, page, perPage) => {
```
timtamimi, line 39: Something is a bit weird here but I'm not sure what it is. Have a look at the screenshot below from the UI
image

This value is coming out blank for me. Am I doing something wrong? Or have we accidentally changed one of the variable names which no longer match between the service and the UI?

timtamimi: Edit: It could just be correct.

timtamimi (approved):
I left a few "advisory comments" which you may or may not want to look at, but I'm happy with this all in all so I'm cool with it if you want to merge it as is. So I've marked it as Approved.

src/lib/connectors/repos/scheduled-notifications.js
james-siteclick: There was a bug where pre-2008 returns were displayed - fixed now

PR #1402, opened 2021-05-17
Feat/water 3136 purpose conditions

timtamimi:
Created on behalf of Stephan for ticket WATER-3136


migrations/sqls/20210517130242-import-licence-version-purpose-conditions-up.sql
```diff
@@ -0,0 +1,31 @@
...
+  "param_1" VARCHAR,
+  "param_2" VARCHAR,
+  "notes" VARCHAR,
+  "external_id" VARCHAR NOT NULL,
+  "date_created" timestamp NOT NULL DEFAULT NOW(),
+  "date_updated" timestamp NOT NULL DEFAULT NOW(),
+    CONSTRAINT "uidx_external_id" unique ("external_id"),
+    CONSTRAINT "licence_version_purpose_id" FOREIGN KEY ("licence_version_purpose_id")
+    REFERENCES "water"."licence_version_purposes" ("licence_version_purpose_id") MATCH SIMPLE
+    ON DELETE CASCADE,
```
james-siteclick, line 26: We haven't previously used `on delete cascade` because we are moving more logic to the DB layer.

So I think we should discuss with other devs and decide whether it should be used in future.

james-siteclick (changes requested):

james-siteclick (approved):

PR #1432, opened 2021-06-07
Water 3105 - delete re-billed invoices

Stephan-Hough:
feat/water-3105
-- delete rebilling invoice licences and transactions added to delete process
-- delete by originalInvoiceId removed from invoice service

### DIFF
--- a/src/modules/billing/services/batch-service.js
+++ b/src/modules/billing/services/batch-service.js
@@ -326,25 +326,13 @@ const updateInvoiceLicencesForSupplementaryReprocessing = invoice => {
  * the linked invoices where the invoice id does not match the original invoice id
  * @param {*} invoiceId
  */
-const deleteTransactionsInvoiceLicencesInvoices = async invoice => {
-  if (invoice.billingInvoiceId !== invoice.originalBillingInvoiceId) {
-    await newRepos.billingTransactions.deleteByInvoiceId(invoice.billingInvoiceId);
-    await newRepos.billingInvoiceLicences.deleteByInvoiceId(invoice.billingInvoiceId);
-    await newRepos.billingInvoices.delete(invoice.billingInvoiceId);
-  }
-};
-
-/**
- * Deletes all rebilled invoices that is grouped by the original invoice id and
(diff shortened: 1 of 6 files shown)


src/modules/billing/services/batch-service.js
```diff
@@ -350,26 +352,19 @@ const deleteBatchInvoice = async (batch, invoiceId) => {
...
-
-    // Delete the invoice if the rebilling state is null otherwise
-    // reset the original invoice and delete the rebilled invoices.
-    if (invoice.rebillingState === null) {
-      await newRepos.billingInvoices.delete(invoiceId);
-    } else {
-      await deleteRebillingInvoices(invoice.originalBillingInvoiceId);
+    const invoicesToDelete = [invoice];
+
+    if (invoice.originalBillingInvoiceId !== null) {
```
james-siteclick, line 361: I wonder if:
```
 if (invoice.rebillingState !== null) {
```
Is the more reliable indicator that it's one of the pair

src/modules/billing/services/batch-service.js
```diff
@@ -350,26 +352,19 @@ const deleteBatchInvoice = async (batch, invoiceId) => {
...
+      await invoiceService.updateInvoice(invoice.originalBillingInvoiceId, { isFlaggedForRebilling: false, originalBillingInvoiceId: null, rebillingState: null });
+      invoicesToDelete.push([...invoice.linkedBillingInvoices].filter(row => row.billingInvoiceId !== row.originalBillingInvoiceId));
```
james-siteclick, line 363: I think this may end up attempting to delete one of the pair twice.  This is because the base invoice is only filtered out from `linkedBillingInvoices` at the mapping layer, and this code is using the repo layer directly.

So I think we possibly need to just overwrite invoicesToDelete:
```
 const isNotOriginalInvoice = row => row.billingInvoiceId !== row.originalBillingInvoiceId);

// ...

     invoicesToDelete = invoice.linkedBillingInvoices.filter(isNotOriginalInvoice);
```

james-siteclick (changes requested):

james-siteclick (approved):

PR #1444, opened 2021-06-11
Adds new endpoint for posting a new linkage

timtamimi:
Original ticket: https://eaflood.atlassian.net/browse/WATER-3114

Goes along with: https://github.com/DEFRA/water-abstraction-tactical-idm/pull/737

### DIFF
--- a/src/lib/services/licence-gauging-stations-service.js
+++ b/src/lib/services/licence-gauging-stations-service.js
@@ -1,0 +1,14 @@
+'use strict';
+const licenceGaugingStationsRepo = require('../connectors/repos/licence-gauging-stations');
+
+const createNewLicenceLink = async (gaugingStationId, licenceId, properties = {
+  licenceVersionPurposeConditionId: null,
+  thresholdUnit: 'mAOD',
+  thresholdValue: 0,
+  abstractionPeriod: null,
+  restrictionType: 'flow',
+  alertType: 'reduce',
+  source: 'wrls'
+}) => await licenceGaugingStationsRepo.create({ gaugingStationId, licenceId, ...properties, date_status_updated: new Date() });
+
+exports.createNewLicenceLink = createNewLicenceLink;
--- a/src/lib/services/licence-gauging-stations-service.js
+++ b/src/lib/services/licence-gauging-stations-service.js
@@ -9,6 +9,6 @@ const createNewLicenceLink = async (gaugingStationId, licenceId, properties = {
   restrictionType: 'flow',
   alertType: 'reduce',
   source: 'wrls'
-}) => licenceGaugingStationsRepo.create({ gaugingStationId, licenceId, ...properties, date_status_updated: new Date() });
+}) => licenceGaugingStationsRepo.create({ gaugingStationId, licenceId, ...properties });
 
 exports.createNewLicenceLink = createNewLicenceLink;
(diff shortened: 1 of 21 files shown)


src/lib/services/licence-gauging-stations-service.js
```diff
@@ -0,0 +1,14 @@
...
+
+const createNewLicenceLink = async (gaugingStationId, licenceId, properties = {
+  licenceVersionPurposeConditionId: null,
+  thresholdUnit: 'mAOD',
+  thresholdValue: 0,
+  abstractionPeriod: null,
+  restrictionType: 'flow',
+  alertType: 'reduce',
+  source: 'wrls'
+}) => licenceGaugingStationsRepo.create({ gaugingStationId, licenceId, ...properties, date_status_updated: new Date() });
```
james-siteclick, line 12: I have a vague recollection we were going to set `date_status_updated` when a notification is sent out.

I wonder then if `date_status_updated` should be `null` until then?

src/lib/services/licence-gauging-stations-service.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+const licenceGaugingStationsRepo = require('../connectors/repos/licence-gauging-stations');
+
+const createNewLicenceLink = async (gaugingStationId, licenceId, properties = {
+  licenceVersionPurposeConditionId: null,
+  thresholdUnit: 'mAOD',
+  thresholdValue: 0,
+  abstractionPeriod: null,
```
james-siteclick, line 8: Does there need to be some code to map the `abstractionPeriod` object above, e.g.:
```
{
  startDay: 1,
  startMonth: 1,
  endDay: 31,
  endMonth: 12
}
```
To the DB fields `abstraction_period_start_day` etc...

src/modules/gauging-stations/controller.js
```diff
@@ -12,17 +15,56 @@ const getGaugingStation = async request => {
...
+    await licencesService.getLicenceById(licenceId);
+
+    // Check that the gauging station ID belongs to an actual gauging station.
+    await gaugingStationService.getGaugingStation(gaugingStationId);
+
+    return licenceGaugingStationsService.createNewLicenceLink(gaugingStationId, licenceId, {
+      licenceVersionPurposeConditionId,
+      thresholdUnit,
+      thresholdValue,
+      ...abstractionPeriod,
```
james-siteclick, line 52: The abstraction period here is destructured whereas the `createNewLicenceLink()` default param indicates it is expecting an object so unclear if this will work.

The Joi validation rules on the route below look correct though.

src/lib/services/licence-gauging-stations-service.js
timtamimi: It's currently not nullable, but if you feel this is a problem we can create a migration script to make it nullable. What do you think?

timtamimi: Discussed with James. This will be resolved in commit d6436849

src/lib/services/licence-gauging-stations-service.js
timtamimi: You're totally right. I will flatten this object and correct the naming mismatch. Thank you for spotting this.

timtamimi: Resolved in 5c9be4f8

src/modules/gauging-stations/controller.js
timtamimi: Flattened and corrected in 5c9be4f8

james-siteclick (approved):

PR #1446, opened 2021-06-14
Feat/water 3206 rebilling cm integration 2

james-siteclick:
`WATER-3206`

* Integrates the CM re-billing into the billing process.
* Invoices flagged for re-billing are sent to the CM during the billing process
* Local WRLS invoices are also created
* Other jobs/services modified to handle re-billing/non-rebilling invoices


src/modules/billing/controllers/invoices.js
```diff
@@ -1,27 +1,14 @@
...
-    if (!invoiceIsPartOfSentBatch(invoice)) {
-      return Boom.conflict('Cannot update invoice that is not part of a sent batch');
-    }
-    if (invoiceIsRebill(invoice)) {
-      return Boom.conflict('Cannot update invoice that is itself a rebill');
-    }
-    return invoiceService.updateInvoice(invoiceId, request.payload);
+    const { isFlaggedForRebilling } = request.payload;
+
+    return await invoiceService.setIsFlaggedForRebilling(invoiceId, isFlaggedForRebilling);
```
timtamimi, line 11: Is `return await...` valid syntax?

I would expect `return [the Promise]` normally

src/modules/billing/jobs/prepare-transactions.js
```diff
@@ -9,14 +10,19 @@ const { BATCH_ERROR_CODE } = require('../../../lib/models/batch');
...
 const { logger } = require('../../../logger');
 const supplementaryBillingService = require('../services/supplementary-billing-service');
 const billingTransactionsRepo = require('../../../lib/connectors/repos/billing-transactions');
+const Transaction = require('../../../lib/models/transaction');
 
 const createMessage = partial(helpers.createMessage, JOB_NAME);
 
 const getTransactionId = transaction => transaction.billingTransactionId;
 
+const isCandidateTransaction = transaction => transaction.status === Transaction.statuses.candidate;
```
timtamimi, line 24: Would you kindly slot in an inline comment explaining what a candidate transaction is please?

src/modules/billing/jobs/rebilling.js
```diff
@@ -49,7 +51,7 @@ const handler = async job => {
 
   // Process each invoice that needs rebilling
   for (const invoice of invoices) {
-    await invoiceService.rebillInvoice(batch, invoice);
+    await rebillingService.rebillInvoice(batch, invoice.id);
```
timtamimi, line 54: Is the await here useful/necessary?

src/modules/billing/mappers/transaction.js
```diff
@@ -96,7 +97,8 @@ const dbToModelMapper = createMapper()
   .map('volume').to('volume', volume => volume ? parseFloat(volume) : null)
```
timtamimi, line 97: If I remember correctly, Russell recently fed back that the it's OK for volumes to be zero by default. With that in mind, should it be `.map('volume').to('volume', volume => volume ? parseFloat(volume) : 0)` as opposed to `.map('volume').to('volume', volume => volume ? parseFloat(volume) : null)`?

Ignore if not

src/modules/billing/mappers/transaction.js
```diff
@@ -96,7 +97,8 @@ const dbToModelMapper = createMapper()
...
   .map().to('agreements', mapDBToAgreements)
   .map(['billingVolume', 'endDate', 'season']).to('billingVolume',
-    (billingVolume, endDate, season) => billingVolume ? getBillingVolumeForTransaction({ billingVolume, endDate, season }) : null);
+    (billingVolume, endDate, season) => billingVolume ? getBillingVolumeForTransaction({ billingVolume, endDate, season }) : null)
```
timtamimi, line 100: Same here with regards to defaulting to null vs defaulting to zero. I'm not sure which is "more correct".

timtamimi (approved):

src/modules/billing/controllers/invoices.js
james-siteclick: Yes it is, it means any error that occurs in the promise will be handled by the try/catch, whereas this is not the case if you return the promise directly

src/modules/billing/jobs/rebilling.js
james-siteclick: It means that the tasks are processed in series (avoids overwhelming our DB/CM) and should avoid uncaught errors

src/modules/billing/mappers/transaction.js
james-siteclick: I think it's OK here because billingVolume is a child object which may or may not be present

src/modules/billing/mappers/transaction.js
james-siteclick: I think it's OK for the volume to be null for min charge transactions.  However I agree the logic looks suspect.  

I might expect it to be:
```
 volume => isNull(volume) ? null : parseFloat(volume)
```
otherwise it won't handle zero properly.

src/lib/services/invoice-service.js
```diff
@@ -14,24 +14,46 @@ const mappers = require('../../modules/billing/mappers');
 const FinancialYear = require('../models/financial-year');
 
 // Errors
-const { NotFoundError } = require('../errors');
+const { NotFoundError, ConflictingDataError } = require('../errors');
 
-const getInvoiceById = async invoiceId => {
+const getInvoiceDataById = async invoiceId => {
```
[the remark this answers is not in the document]
james-siteclick: Just refactoring / reducing duplication.  The implementation/naming of `getInvoiceById` is unchanged, but it now calls this (as does something else).

src/lib/models/model.js
```diff
@@ -73,6 +73,11 @@ class Model {
 
     return momentDate;
   };
+
+  clearId () {
```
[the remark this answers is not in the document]
james-siteclick: Done

src/modules/billing/jobs/prepare-transactions.js
james-siteclick: Added to the service model

PR #1461, opened 2021-07-02
Minor tweak following QA identifying that previous implementation was…

timtamimi:
… hard-deleting records

Relates to: https://eaflood.atlassian.net/browse/WATER-3116


uijim (approved):
I can't test this at the moment but the code change looks good to me. 👍🏻

PR #1464, opened 2021-07-12
Additional field added for WATER-3117

DashCoder:


src/lib/connectors/repos/queries/gauging-stations.js
```diff
@@ -1,4 +1,5 @@
 exports.findGaugingStationWithLinkedLicences = `select
+lgs.licence_gauging_station_id,
```
timtamimi, line 2: Please fix the indentation so that it's consistent with the following lines.

src/lib/connectors/repos/queries/gauging-stations.js
```diff
@@ -33,6 +34,7 @@ and lgs.date_deleted is null
 and gs.gauging_station_id=:gaugingStationId
 union all
 select
+lgs.licence_gauging_station_id,
```
timtamimi, line 37: Please note you can also get this from `gs.gauging_station_id` - same outcome but slightly less computational power.

timtamimi (changes requested):

src/lib/connectors/repos/queries/gauging-stations.js
timtamimi: Ignore this - I was mistaken.

timtamimi (approved):

PR #1467, opened 2021-07-14
WATER-3232 - Fixtures loader extended and due and completed returns data added

Stephan-Hough:
feat/water-3232
-- fixtures loader extended to receive custom reference data
-- returns loader now calculates future return due dates
-- set loader adds refs to existing fixtures loader instance

### DIFF
--- a/integration-tests/billing/fixtures/test-returns.yaml
+++ b/integration-tests/billing/fixtures/test-returns.yaml
@@ -135,7 +135,7 @@
     licence_type: abstraction
     start_date: '2021-01-01'
     end_date: '2021-02-28'
-    due_date:	'2021-03-31'
+    due_date:	$returnDates.dueDate
     returns_frequency: month
     status: due
     source: NALD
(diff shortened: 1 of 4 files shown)


integration-tests/billing/fixtures/test-returns.yaml
```diff
@@ -253,3 +253,66 @@
...
+            }
+        ],
+        "isCurrent":true,
+        "description":"Its all about the description",
+        "isTwoPartTariff":true
+      }
+    received_date: null
+    return_requirement: '9999992'
+    under_query: false
+    is_test: true
```
timtamimi, line 318: Please insert a blank line at the bottom. Yaml files don't get automatically linted.

integration-tests/billing/services/returns-loader.js
```diff
@@ -18,13 +19,21 @@ const createConnector = tail => async body => {
 };
 
 const create = () => {
+  // create a future date and pass it to the fixture loader as reference for use in yaml objects
+  const refDates = {
+    name: '$returnDates',
+    obj: {
+      dueDate: moment().add(1, 'month').format('YYYY-MM-DD')
```
timtamimi, line 26: This is fine, as long as we don't care about the actual due date...(?)

timtamimi (approved):

integration-tests/billing/services/returns-loader.js
Stephan-Hough: Thanks - yes I added an additional return specifically for this test scenario to ensure the status will always show as due on the ui. If the due date needs to be tested then the same logic could be used in Cypress or an alternative return with a static due date could be checked.

PR #1470, opened 2021-07-20
BUGFIX - WATER 3278 - 2PT duplicate charges

Stephan-Hough:
bugfix/water-3278
--get 2pt batches in connectors now includes supplementary batches because you can have 2PT invoices in supplementary batches.


uijim (approved):
Looks good

PR #1489, opened 2021-08-05
update billingbatches with creditNoteCount

DashCoder:
Dashboard to count number of creditbills


timtamimi (changes requested):
This is partially correct.

What you did in batch-service.js is spot on +1

But you were incorrect to change the validator to allow positive integers. Instead, we want to ensure that the incoming figure is always negative. 

Please review my PR instead: https://github.com/DEFRA/water-abstraction-service/pull/1492

PR #1490, opened 2021-08-06
Bugfix - water - 3276 - incorrect historic transactions

Stephan-Hough:
bugfix/water-3276
-- added attribute to transaction class
to get the charge element purpose use code
for generating the transaction hash during a bill run


src/modules/billing/services/supplementary-billing-service/supplementary-processor.js
```diff
@@ -73,7 +73,7 @@ const getGroupingKey = transaction => {
   // Get the data and serialize to a JSON string
   const data = {
     ...pick(transaction, keys),
-    description: transaction.description.trim().toLowerCase(),
+    description: transaction.chargeElementPurposeUseCode,
```
timtamimi, line 76: Would you mind changing the `description` field here to another name, such as `chargeElementPurposeUseCode` - just since this isn't actually pulling the description any more

timtamimi (approved):

PR #1501, opened 2021-08-20
Bugfix - water -3245 - rebilling

Stephan-Hough:
bugfix/water-3245
-- rebilling job update process of original invoice fixed


DashCoder (approved):

DashCoder (changes requested):
Code seems to be working, however may benefit from refactoring to move logic from UI to Backend.

PR #1508, opened 2021-08-26
route and controller for clearing rebilling flag by batchid

DashCoder:


Stephan-Hough (dismissed):

src/modules/billing/routes/invoices.js
```diff
@@ -23,3 +23,19 @@ exports.patchInvoice = {
     }
   }
 };
+
+exports.resetIsFlaggedForRebilling = {
+  method: 'PATCH',
+  path: '/water/1.0/billing/invoices/rebillingflag/{batchId}',
```
Stephan-Hough, line 29: You need to add the invoiceId as another parameter to only reset the flag for that specific invoice

src/modules/billing/controllers/invoices.js
```diff
@@ -14,4 +14,15 @@ const patchInvoice = async request => {
   }
 };
 
+const resetIsFlaggedForRebilling = async request => {
+  try {
+    const { batchId } = request.params;
+
+    return await invoiceService.resetIsFlaggedForRebilling(batchId);
```
Stephan-Hough, line 21: You will need to add another method to only cancel the reissue of a specific invoice. This will use the invoice id received form the route

Stephan-Hough (changes requested):
See the comments added. You will need to update the route to receive the invoice id and a new invoice service method to only update the rebilling status for that one invoice using the invoice id received from the route parameter

src/modules/billing/routes/batches.js
```diff
@@ -125,13 +125,15 @@ const getBatchInvoiceDetail = {
...
+  path: `${BASE_PATH}/{batchId}/invoices/{invoiceId}/org/{originalInvoiceId}/rebill/{rebillInvoiceId}`,
   handler: controller.deleteBatchInvoice,
   config: {
     validate: {
       params: Joi.object().keys({
         batchId: Joi.string().uuid().required(),
-        invoiceId: Joi.string().uuid().required()
+        invoiceId: Joi.string().uuid().required(),
+        originalInvoiceId: Joi.string().uuid(),
+        rebillInvoiceId: Joi.string().uuid()
```
Stephan-Hough, line 136: originalInvoiceId and rebillInvoiceId should be added as Joi.string().uuid().optional() query parameters.

src/modules/billing/routes/invoices.js
```diff
@@ -23,3 +23,21 @@ exports.patchInvoice = {
     }
   }
 };
+
+exports.resetIsFlaggedForRebilling = {
+  method: 'PATCH',
+  path: '/water/1.0/billing/invoices/rebillingflag/{batchId}/org/{originalInvoiceId}/rebill/{rebillInvoiceId}',
```
Stephan-Hough, line 29: the route here receives the batch id but the controller doesn't use it so this can be removed. i would suggest a route of: 
`/water/1.0/billing/invoices/{invoiceId}/reset-rebilling-flag`

the ui will then send the original invoice id to this route to set the status to rebill when the status is rebilled or null if the status is rebill

src/modules/billing/controllers/invoices.js
```diff
@@ -14,4 +14,25 @@ const patchInvoice = async request => {
...
+  } catch (err) {
+    return mapErrorResponse(err);
+  }
+};
+
+const resetIsFlaggedForRebillingByInvoiceId = async request => {
+  try {
+    const { originalInvoiceId } = request.params;
```
Stephan-Hough, line 29: this will just be the invoiceId instead of originalInvoiceId because within this context we only receive on id

src/lib/connectors/repos/queries/billing-invoices.js
```diff
@@ -47,3 +47,9 @@ where
   and t2.original_billing_invoice_id is not null
 returning *;
 `;
+
+exports.resetIsFlaggedForRebillingByInvoiceId = `
```
Stephan-Hough, line 51: you will also need to update the rebilling state for the invoice as well as the original invoice id column.

If the original_invoice_id === invoice_id then this means it is the first rebill in a rebilling chain then set rebilling_status = null and original_invoice_id = null
otherwise set the rebilling_state = 'rebill' and leave the original_invoice_id as is

src/modules/billing/controllers/invoices.js
```diff
@@ -14,4 +14,25 @@ const patchInvoice = async request => {
   }
 };
 
+const resetIsFlaggedForRebilling = async request => {
```
Stephan-Hough, line 17: Sorry if I didn't explain this better after my first review. The resetting of the rebilling status for the original invoice id is already handled on line 115 in the delete invoice controller so you don't need these controller functions here and you don;t need the routes for these either.

Stephan-Hough (changes requested):
The back end process for canceling the rebill of an invoice requires 3 steps to be executed. These steps is executed through the existing delete invoice route that has optional query params for originalInvoiceId and rebillingInvoiceId. 

when the originalInvoiceId and rebillingInvoiceId is not null then;
1. Delete the rebill invoice using invoiceId
2. Delete the reversal invoice using the rebillingInvoiceId
3. Update the rebilling_state and original_invoice_id  using the originalInvoiceId

src/modules/billing/controllers/batches.js
```diff
@@ -110,9 +110,21 @@ const getBatchInvoiceDetail = async request => {
...
   const { batch } = request.pre;
   const { invoiceId } = request.params;
+
+  let originalInvoiceId = null;
+  let rebillInvoiceId = null;
+
+  if (request.query) {
+    originalInvoiceId = request.query.originalInvoiceId;
+    rebillInvoiceId = request.query.rebillInvoiceId;
+  }
```
Stephan-Hough, line 120: This could be written as 
```const { originalInvoiceId, rebillInvoiceId  } = request.query;```

and then you check if there are values for these on line 122

File: src/modules/billing/controllers/batches.js [hunk 9]
```diff
@@ -110,9 +110,21 @@ const getBatchInvoiceDetail = async request => {
...
   try {
-    // Delete the invoice
-    await batchService.deleteBatchInvoice(batch, invoiceId);
+    if (originalInvoiceId && rebillInvoiceId) {
+      await batchService.deleteBatchInvoice(batch, invoiceId, originalInvoiceId, rebillInvoiceId); /* delete both related bills */
+    } else {
+      // Delete the invoice
+      await batchService.deleteBatchInvoice(batch, invoiceId);
+    }
```
Stephan-Hough, line 127: You don't need the if statement here if you are checking the values exists in the batch-service. 

You could just leave this line below and get rid of the rest

```await batchService.deleteBatchInvoice(batch, invoiceId, originalInvoiceId, rebillInvoiceId);```

timtamimi (approved):

PR #1527, opened 2021-10-07
Bugfix - water 3338 - incorrect Two Part Tariff 2nd part charges for historic invoices

Stephan-Hough:
bugfix/water-3338
-- reverting previous fix for 3323 - this is causing regression issues
-- successfully retested 3323 with this change

### DIFF
--- a/test/modules/billing/services/transactions-service.js
+++ b/test/modules/billing/services/transactions-service.js
@@ -15,7 +15,7 @@ const billingVolumesService = require('../../../../src/modules/billing/services/
 const repos = require('../../../../src/lib/connectors/repos');
 const { logger } = require('../../../../src/logger');
 
-const { createTransaction, createInvoiceLicence, createTransactionDBRow, createBatch, createInvoice, createChargeVersionYear } = require('../test-data/test-billing-data');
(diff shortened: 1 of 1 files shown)


src/modules/billing/services/transactions-service.js
```diff
@@ -91,6 +91,19 @@ const getBatchTransactions = batch => flatMap(batch.invoices.map(getInvoiceTrans
...
 const getTransactionId = transaction => transaction.id;
 
+const getBatchTransactionHistory = async batchId => {
+  const historicTransactions = await newRepos.billingTransactions.findHistoryByBatchId(batchId);
+
+  // get licences that need to have the 2nd part recalculated for supplementary
+  const twoPartTariffChargeVersionYears = await newRepos.billingBatchChargeVersionYears.findTwoPartTariffByBatchId(batchId);
+
+  // filter the transaction so that it does not include any 2PT transactions where there is no 2PT charge version year
+  return historicTransactions.filter(trx =>
```
timtamimi, line 101: This works, so it's "fine" but does look a bit messy. To get rid of this **in the future** I would like us to slot a flag in the DB that tells us whether a given transaction is a first part, a second part or neither.

PR #1537, opened 2021-10-25
WATER-3347: Migrate the returns module PGBoss jobs to BullMQ

defra-contractor:
See: https://eaflood.atlassian.net/browse/WATER-3347 as this covers everything required in the ticket now.


src/modules/returns-notifications/lib/returns-notification-send.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+  ];
+};
+
+const handleReturnsNotificationSend = async job => {
+  logger.info(`Handling: ${JOB_NAME}:${job.id}`);
+};
+
+const onComplete = async (job, queueManager) => {
+  const options = await send.prepareMessageData(job.data);
```
Stephan-Hough, line 28: Why did you not add these steps in the handler? Is it not better to perform the job functions in the handler rather than complete?

src/modules/returns-notifications/lib/returns-notification-send.js
```diff
@@ -0,0 +1,39 @@
...
+  await notify.enqueue(queueManager, options);
+  logger.info(`${JOB_NAME}: Job has completed`);
```
Stephan-Hough, line 30: Why did you not add these steps in the handler? Is it not better to perform the job functions in the handler rather than complete?

src/modules/returns/lib/jobs/map-to-json.js
```diff
@@ -87,6 +95,19 @@ const uploadJsonToS3 = (eventId, json) => {
...
 
+const onFailed = async (job, err) => {
+  logger.error(`${JOB_NAME}: Job has failed`, err);
+};
+
+const onComplete = async (job, queueManager) => {
+  // Format and add BullMQ message
+  const { eventId, companyId } = job.data;
+  await queueManager.add(validateReturns.jobName, { eventId, companyId });
+  logger.info(`${JOB_NAME}: Job has completed`);
```
Stephan-Hough, line 106: As above, Why did you not add these steps in the handler? Is it not better to perform the job functions in the handler rather than complete?

Stephan-Hough, line 106: As above, Why did you not add these steps in the handler? Is it not better to perform the job functions in the handler rather than complete?

Stephan-Hough (changes requested):
I have added a couple of questions about the steps in the onComplete part of the jobs, not sure if that would be better placed in the handler instead

src/modules/returns-notifications/lib/returns-notification-send.js
defra-contractor: I tried to do that but didn't have access to the queueManager from the handler which is why I moved into the onComplete function.

src/modules/returns/lib/jobs/map-to-json.js
defra-contractor: Same reason as above and also the validateReturns job is supposed happen on the successful completion of the map-to-json job.

Stephan-Hough: thanks

PR #1565, opened 2021-11-11
bugfix/water-3376

Stephan-Hough:
-- transaction model validation added for volume to not be greater than
the max annual quantity


timtamimi (approved):
Code changes look fine - But I would recommend running a local test please first before handing over to QA :D Many thanks.

PR #1577, opened 2021-11-18
Water - 3373 - isTwoPartSecondPartCharge transaction flag

Stephan-Hough:
techdebt/water-3373
isTwoPartSecondPartCharge flag replaces old isTwoPartSupplementary flag

### DIFF
--- a/migrations/sqls/20211118092400-second-part--charge-transaction-flag-down.sql
+++ b/migrations/sqls/20211118092400-second-part--charge-transaction-flag-down.sql
@@ -8,4 +8,4 @@ add column is_two_part_tariff_supplementary boolean default false not null;
 
 update water.billing_transactions 
 set is_two_part_tariff_supplementary = true
-where lower(description) like 'second%';
+where description ilike 'second%';
--- a/migrations/sqls/20211118092400-second-part--charge-transaction-flag-down.sql
+++ b/migrations/sqls/20211118092400-second-part--charge-transaction-flag-down.sql
@@ -1,0 +1,11 @@
+alter table water.billing_transactions 
+drop column is_two_part_second_part_charge;
+
+
+alter table water.billing_transactions 
+add column is_two_part_tariff_supplementary boolean default false not null;
+
+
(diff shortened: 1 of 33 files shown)


migrations/sqls/20211118092400-second-part--charge-transaction-flag-down.sql
```diff
@@ -0,0 +1,11 @@
...
+drop column is_two_part_second_part_charge;
+
+
+alter table water.billing_transactions 
+add column is_two_part_tariff_supplementary boolean default false not null;
+
+
+update water.billing_transactions 
+set is_two_part_tariff_supplementary = true
+where lower(description) like 'second%';
```
timtamimi, line 11: this is absolutely fine but you could also do:

```
where description ilike 'second%';
```

`ilike` does a case-insensitive comparison.

src/modules/billing/services/cm-refresh-service.js
```diff
@@ -106,8 +106,7 @@ const mapTransaction = (transactionMap, cmTransaction) => {
     : new Transaction();
 
   return transaction.fromHash({
-    ...transactionMapper.cmToPojo(cmTransaction),
-    isTwoPartTariffSupplementary: false
+    ...transactionMapper.cmToPojo(cmTransaction)
```
timtamimi, line 109: No need to itinerate through object since we're only returning a single thing.

Consider:
```
return transaction.fromHash(transactionMapper.cmToPojo(cmTransaction));
```

timtamimi (approved):

migrations/sqls/20211118092400-second-part--charge-transaction-flag-down.sql
Stephan-Hough: ooo nice one thanks didn't know this

src/modules/billing/services/cm-refresh-service.js
Stephan-Hough: good catch thanks

PR #1596, opened 2021-12-22
WATE - 3398 SROC charge category lookup service

Stephan-Hough:
Feat/water 3398
-- charge category lookup end point controller and service added


src/modules/billing/jobs/sync-charge-categories.js
```diff
@@ -11,7 +11,7 @@ const s3Connector = require('../../../lib/services/s3');
 
 // Constants
 const JOB_NAME = 'billing.charge-categories.sync-from-csv';
-const csvKey = 'billing-metadata/charge-categories.csv';
+const csvKey = 'billing-metadata/charge-categories-v1.csv';
```
timtamimi, line 14: Why `-v1`?

src/modules/charge-categories/routes.js
```diff
@@ -0,0 +1,24 @@
...
+  getChargeCategoryByProperties: {
+    method: 'GET',
+    path: '/water/1.0/charge-categories',
+    handler: controller.getChargeCategoryByProperties,
+    config: {
+      validate: {
+        query: Joi.object().keys({
+          source: Joi.string().required().valid('tidal', 'non-tidal'),
+          loss: Joi.string().required().valid('low', 'medium', 'high'),
+          waterAvailability: Joi.string().required().valid('restricted availablity or no availability', 'available'),
```
timtamimi, line 16: Is there any reason why `waterAvailability` isn't simply a boolean?

test/modules/charge-categories/controller.js
```diff
@@ -0,0 +1,61 @@
...
+experiment('./src/modules/change-reasons/controller.js', () => {
+  beforeEach(async () => {
+    sandbox.stub(repos.chargeCategories, 'findOneByProperties').resolves(chargeCategories);
+  });
+
+  afterEach(async () => {
+    sandbox.restore();
+  });
+
+  experiment('.getChargeCategoryByProperties', () => {
```
timtamimi, line 36: I would _recommend_ adding a couple of other tests. (Not mandatory - just a suggestion)

E.g. Breach maxVolume

E.g. What happens if some values are missing from the payload.

E.g. Does it identify the **right** Charge Category given a set of inputs, from an array of available charge categories

timtamimi (changes requested):

src/modules/billing/jobs/sync-charge-categories.js
Stephan-Hough: removed the -v1 was used for testing locally without upsetting dev

src/modules/charge-categories/routes.js
Stephan-Hough: changed to isRestrictedSource and aligned with UI

test/modules/charge-categories/controller.js
Stephan-Hough: Those are really good points - the database is stubbed in these tests so will make it a little inaccurate to test at this level but will flag this in the ticket to test manually. 

The route defines the parameter constraints and validates these and I have added tests for this to ensure the correct payload is received. 

and the last point is probably again a manual test because the database is stubbed for these tests.

PR #1616, opened 2022-02-03
WATER-3478

mo-zag:
Removed unused charging module statuses.


timtamimi:
This shouldn't be merged until CM have deployed v0.16 or v0.17 onto INT and PRD CM.

Stephan-Hough (approved):

PR #1620, opened 2022-02-07
bugfix/cv-start-and-end-dates-the-same

Stephan-Hough:
do not select charge versions where the start and end dates are the same

### DIFF
--- a/src/lib/connectors/repos/queries/charge-versions.js
+++ b/src/lib/connectors/repos/queries/charge-versions.js
@@ -46,6 +46,7 @@ from (
     on cv.change_reason_id = cr.change_reason_id
   where l.region_id=:regionId
     and cv.status='current'
+    and cv.start_date <> cv.end_date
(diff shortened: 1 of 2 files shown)


src/lib/connectors/repos/queries/charge-versions.js
```diff
@@ -111,6 +112,8 @@ from (
     on cv.change_reason_id = cr.change_reason_id
   where l.region_id=:regionId
     and cv.status='current'
+    and cv.start_date <> cv.end_date
```
timtamimi, line 115: Duplicate lines

timtamimi (changes requested):

src/lib/connectors/repos/queries/charge-versions.js
Stephan-Hough: removed

timtamimi (approved):

PR #1624, opened 2022-02-10
bugfix/missing-credits

Stephan-Hough:
-- do not create charge version workflows for
licences in an active bill run


timtamimi:
This is fine but could we/should we rename the query from `findIdsCreatedAfterDate` to something more meaningful?

I note that the method above it is called `findIdsByDateNotInChargeVersionWorkflows` which potentially also needs to be renamed.

How about `getNewLicenceVersionsForChargeVersionWorkflow` or something to that effect?

I'll pre-emptively approve the PR anyway

timtamimi (approved):

PR #1636, opened 2022-02-22
WATER-3415: Addition of Notes to Charge Information review page

defra-contractor:
See: https://eaflood.atlassian.net/browse/WATER-3415

This is a prerequisite for: https://github.com/DEFRA/water-abstraction-ui/pull/2048

### DIFF
--- a/migrations/sqls/20220224164015-create-note-table-up.sql
+++ b/migrations/sqls/20220224164015-create-note-table-up.sql
@@ -1,0 +1,19 @@
+create table water.notes (
+    note_id uuid primary key default public.gen_random_uuid(),
+    text varchar(500) not null,
+    date_created timestamp not null default now(),
+    date_updated timestamp not null default now(),
+    user_id integer not null,
+    type varchar not null,
+    type_id uuid not null,
+    licence_id uuid null references licences(licence_id)
+);
+
+alter table water.charge_versions add column note_id uuid null;
+
+alter table water.charge_versions
+    add constraint charge_version_note_id_fkey
+        foreign key (note_id)
+            references water.notes (note_id)
+                match simple
+            on delete cascade;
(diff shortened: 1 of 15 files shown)


src/lib/models/note.js
```diff
@@ -0,0 +1,59 @@
...
+  get licenceId () { return this._licenceId; }
+
+  set licenceId (licenceId) {
+    validators.assertNullableId(licenceId);
+    this._licenceId = licenceId;
+  }
+
+  get userId () { return this._userId; }
+
+  set userId (userId) {
```
Stephan-Hough, line 52: we could change this to be a user model instead of just the userId - but feasible to leave as is and extend later

migrations/sqls/20220224164015-create-note-table-up.sql
```diff
@@ -0,0 +1,20 @@
+drop table if exists water.notes;
```
Stephan-Hough, line 1: doesn't make any difference if this is here or not so we can remove it if you want

src/lib/connectors/repos/notes.js
```diff
@@ -0,0 +1,26 @@
...
+  Note
+    .forge({ noteId })
+    .save(changes);
+
+/**
+ * Find single note by ID
+ * @param {String} noteId
+ * @return {Promise<Object>}
+ */
+const findOne = async noteId => helpers.findOne(Note, 'noteId', noteId, withRelated);
```
Stephan-Hough, line 22: with related will not be used but again it can be left for future use

Stephan-Hough (approved):
Approved with some minor optional changes if you agree

src/lib/models/note.js
defra-contractor: I did think about this and thought I should just go with id for now.  Maybe I'll leave so we can get this change in.

src/lib/connectors/repos/notes.js
defra-contractor: I might leave

PR #1638, opened 2022-02-28
WIP

timtamimi:
Relates to https://eaflood.atlassian.net/browse/WATER-3485
Goes with https://github.com/DEFRA/water-abstraction-ui/pull/2051


Stephan-Hough (changes requested):
Should we not just remove the database columns for source and season and then also remove those from the models, mappers and charge purpose service otherwise it will leave a lot of redundant code. Happy to discuss

Stephan-Hough:
> > Should we not just remove the database columns for source and season and then also remove those from the models, mappers and charge purpose service otherwise it will leave a lot of redundant code. Happy to discuss
> 
> Entirely up to you mate! :D I left them in because they are still in use by ALCS charge versions.
> 
> When we create an ALCS CV, we're still populating those columns. That's why I left them in and made them nullable.
> 
> Happy to remove them if you feel it is safe for us to do so. Let me know.

Yeah it will be safe to remove the source and season from the charge purpose db table mappers, models and charge purpose service. We don't use charge purposes for ALCS charge versions. The source and season is in the charge element model, mappers for charge elements. Cheers

Stephan-Hough (approved):

PR #1644, opened 2022-03-07
Ci/multiple jobs pipeline

jonathangoulding:
Break CI Pipeline into multiple jobs

Add matrix to establish any breaks from upgrading from node v12 to v14 (can extend to v16 in the matrix)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: CI
 
 on:
   workflow_dispatch:
+    tags:
+      description: 'Manual Workflow trigger'
   push:
     branches:
       - develop
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,111 +1,97 @@
-name: CI
-
-on:
-  workflow_dispatch:
-  push:
-    branches:
-      - develop
-  pull_request:
-    types: [ opened, synchronize, reopened ]
-
-jobs:
-  base-node-pipeline: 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -3,14 +3,17 @@ name: CI
 on:
   push:
     branches:
-      - develop
+      - CI/multiple-jobs-pipeline
```
jonathangoulding, line 6: Put back as develop. 

propose CI on all branches - inc feature branches

.github/workflows/ci.yml
```diff
@@ -1,16 +1,22 @@
...
   pull_request:
     types: [ opened, synchronize, reopened ]
 
 jobs:
-  build:
-    # You must use a Linux environment when using service containers or container jobs
+  base-node-pipeline: 
+    uses: DEFRA/water-abstraction-orchestration/.github/workflows/base-node-pipeline.yml@develop
```
timtamimi, line 13: This is clever. 

Presumably we can do the same across the other repos (UI/Returns/Permits/Import) etc? :D

jonathangoulding: All should / can share the common pipeline. 

Best to discuss as a team on the common jobs and what node versions to test

jonathangoulding: test setup looks a little different per repo so keep the test ci as a job in each repo.

jonathangoulding:
Currently allowing audit to ` continue-on-error: true`

Job added to highlight potential vulnerabilities and issues with packages

PR #1662, opened 2022-04-04
Feat/water 3602

Stephan-Hough:


config.js
```diff
@@ -41,7 +41,8 @@ module.exports = {
     returnsGracePeriod: process.env.RETURNS_GRACE_PERIOD || 21,
     // The date SROC goes live
     isSrocLive,
-    srocStartDate
+    srocStartDate,
+    alcsEndYear: 2022
```
jonathangoulding, line 45: no need for configuration from an ENV ?

src/modules/billing/services/batch-service.js
```diff
@@ -254,6 +254,16 @@ const getErrMsgForBatchErr = (batch, regionId) =>
  * @return {Promise<Batch>} resolves with Batch service model
  */
 const create = async (regionId, batchType, toFinancialYearEnding, isSummer) => {
+  let fromFinancialYearEnding, scheme;
+  if (batchType === 'supplementary') {
```
jonathangoulding, line 258: scheme will always be 

```
  scheme = 'alcs';
```

jonathangoulding: Suggestion: 

```
let fromFinancialYearEnding = config.billing.alcsEndYear;
let scheme = 'alcs';

if (batchType === 'supplementary') {
fromFinancialYearEnding = config.billing.alcsEndYear - (config.billing.supplementaryYears - (toFinancialYearEnding - config.billing.alcsEndYear));
}
```

jonathangoulding (commented):
Just a suggestion. Ignore if justified.

config.js
Stephan-Hough: Yeah I guess it could have gone in the env but if you look at the config file all our feature flag and configurations are there. ENV is mainly for secrets and keys

src/modules/billing/services/batch-service.js
Stephan-Hough: Thanks yeah that does look silly

jonathangoulding (approved):
comments addressed

PR #1676, opened 2022-04-25
WATER-3526: Copy note information

defra-contractor:
See: https://eaflood.atlassian.net/browse/WATER-3526

This also requires: https://github.com/DEFRA/water-abstraction-ui/pull/2098

### DIFF
--- a/src/lib/services/charge-versions.js
+++ b/src/lib/services/charge-versions.js
@@ -13,6 +13,7 @@ const noteMapper = require('../mappers/note');
 
 // Services
 const service = require('./service');
+const notesService = require('./notes-service');
 const chargeElementsService = require('./charge-elements');
 const invoiceAccountsService = require('./invoice-accounts-service');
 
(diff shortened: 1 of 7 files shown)


src/lib/services/charge-versions.js
```diff
@@ -65,11 +66,14 @@ const getByLicenceRef = async licenceRef =>
  *
  * @param {String} licenceId The licence id
  */
-const getByLicenceId = async licenceId => service.findMany(
-  licenceId,
-  chargeVersionRepo.findByLicenceId,
-  chargeVersionMapper
-);
+const getByLicenceId = async licenceId => {
```
Stephan-Hough, line 69: I think it would be good if we could add an `optional include notes flag` with default to false, to only return line 75 if set to true otherwise return without the notes. We do not always want to get the note i.e. for billing or on the view licence page you don't need to get the notes so might slow it down in those areas especially for billing.

Stephan-Hough (changes requested):
just one minor comment on the getByLicenceId function on the charge version service

Stephan-Hough (approved):

PR #1682, opened 2022-04-29
fix: billing contact account name

jonathangoulding:
WATER 3644

Theres are multiple names associated with companies this work will introduce a new flow to determine what name to show (just an if else in the UI )

```
if invoice.faoContact 
elif invoice.billingContact.agentCompanyName 
elif invoice.billingContact.roleContact.licenceHolder
elif invoice.billingContact.roleContact.billing
else invoice.name
```

### DIFF
--- a/src/lib/services/invoice-service.js
+++ b/src/lib/services/invoice-service.js
@@ -119,7 +119,8 @@ const decorateInvoiceWithCRMData = (invoice, context) => {
     'address',
     'invoiceAccount',
     'agentCompany',
-    'contact'
+    'contact',
+    'company'
   );
 
   return invoice.fromHash(pickBy(properties));
@@ -143,6 +144,7 @@ const mapToInvoices = async context => {
     const mappedInvoice = await mapInvoice(billingInvoice, context);
     invoices.push(mappedInvoice);
   }
+
   return invoices;
 };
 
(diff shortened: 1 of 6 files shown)


src/lib/services/invoice-service.js
```diff
@@ -200,21 +202,20 @@ const getInvoicesForBatchDownload = async batch => {
...
 
-  const context = { batch, options: Object.assign(defaults, options) };
+  let context = { batch, options: Object.assign(defaults, options) };
 
-  return pWaterfall([
-    getBatchInvoices,
-    getCRMData,
-    mapToInvoices
-  ], context);
+  context = await getBatchInvoices(context);
```
jonathangoulding, line 215: This is the start of an effort to actively remove: 'p-waterfall'

My understanding may be wrong but is is pointless and adds another layer of complexity where it is not needed ?

jonathangoulding: Kept the context object for now. But i would look to isolate the  return objects from function calls in the future.

src/modules/billing/mappers/api/invoice.js
```diff
@@ -9,6 +9,19 @@ const containsWaterUndertaker = invoiceLicences => {
     .includes(true);
 };
 
+const nullToEmpty = (val) => val === null ? '' : val;
+const buildName = (name) => `${nullToEmpty(name.salutation)} ${nullToEmpty(name.firstName)} ${nullToEmpty(name.middleInitials)} ${nullToEmpty(name.lastName)}`;
```
jonathangoulding, line 13: This can generate spaces where they are not needed. 

But this is managed during the page render with `title case`

jonathangoulding (commented):
Some notes for the PR

Cruikshanks (approved):

PR #1683, opened 2022-05-03
Test data fix

Stephan-Hough:
-- changes southern test region  nald region id to 9 to avoid conflicts with the import process using the same nald region id.


Jozzey (approved):
👍🏼

PR #1684, opened 2022-05-11
Bugfix - bulk returns upload

Stephan-Hough:
-- validation repetition removed
-- json file now uploaded to S3 saves validation values
-- tests fixed
-- bull mq job lock increased on all return upload jobs


StuAA78 (approved):
👍

PR #1695, opened 2022-05-17
Add simplified endpoints to retrieve the data relating to batches for a financial year

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-3679

By default when creating a 2-part tariff bill run the current financial year is built. The users have identified scenarios where previous financial years have not been billed. This means they currently cannot generate the bill runs for them.

This change adds a new endpoint that when passed a batch (bill run) type, region, season and current financial will return an array of `unsent` financial years.

It queries the DB for batches that have been created (ignoring, for example, errored ones). It also generates a list of financial years counting 3 back from the current financial year. Any in the second list that appears in the first is removed.

For example

- `2020` exists and has been sent
- `2022` is the current financial year
- `[2020]` is compared to `[2022, 2021, 2020]`
- result sent back is `{ unsentYears: [2022, 2021] }`

This is used in [water-abstaction-ui](https://github.com/DEFRA/water-abstraction-ui/pull/2104) to determine whether to show a page to a user to select the year to bill, and to provide the options to display to the user when the page is shown.

### DIFF
--- a/src/modules/billing/controllers/batches.js
+++ b/src/modules/billing/controllers/batches.js
@@ -25,6 +25,7 @@ const { jobName: approveBatchJobName } = require('../jobs/approve-batch');
 const batchStatus = require('../jobs/lib/batch-status');
 const { BATCH_STATUS } = require('../../../lib/models/batch');
 const { BillingBatch } = require('../../../lib/connectors/bookshelf');
+const { returns, charging } = require('@envage/water-abstraction-helpers');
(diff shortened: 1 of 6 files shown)


Cruikshanks (approved):

StuAA78 (approved):

Jozzey (approved):
Looks fantastic 👍🏼

PR #1696, opened 2022-05-19
chore: use standard js for linting

jonathangoulding:


Jozzey (approved):
Looks good 👍🏼

package.json
```diff
@@ -16,8 +16,7 @@
     "migrate:local-test": "DATABASE_URL=postgres://water_user:password@localhost:5432/permits node scripts/create-schema && DATABASE_URL=postgres://water_user:password@localhost:5432/permits db-migrate up --verbose && DATABASE_URL=postgres://water_user:password@localhost:5432/permits node scripts/migrate-task-config",
     "migrate:down": "db-migrate down --verbose",
     "migrate:create": "db-migrate create --sql-file --",
-    "lint": "eslint .",
-    "lint:fix": "eslint . --fix",
+    "lint": "standard ",
```
StuAA78, line 19: There's a space after `standard` that doesn't need to be there. 

We could keep `lint:fix` as standard does support this

```suggestion
    "lint": "standard",
    "lint:fix": "standard --fix",
```

StuAA78 (changes requested):
Just a couple of small suggested changes to `package.json`

StuAA78 (approved):

PR #1699, opened 2022-05-26
Fix Test Region ID in test data

Cruikshanks:
It was flagged by @jozzey when first building their local environment that NALD imports weren't working. It was tracked down to the Test Region that was being created having a duplicate ID to an existing real region.

We thought we had fixed the problem in [Test data fix](https://github.com/DEFRA/water-abstraction-service/pull/1683) but turns out that was for something else.

This change goes through all the places in the test fixtures where we create the **Test Region** and ensures the ID is set to `9`.


StuAA78 (approved):
👍

PR #1700, opened 2022-05-31
Rename unit test files

StuAA78:
https://github.com/DEFRA/water-abstraction-team/issues/27

Currently, when using VS Code to search through our codebase it's hard to identify which files are code and which are tests. This can be remedied by renaming the test files from `*.js` to `*.test.js`.

This PR renames the unit test files only. The integration tests will be handled in a future PR.


Cruikshanks (approved):
❤️

PR #1705, opened 2022-06-28
Export bill run CSV file for sroc

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3586

We are updating the "download bill run" functionality to support sroc, which entails checking the bill run scheme and using a different set of output fields for alcs and sroc.


Jozzey:
Scheme exists on a batch that is got from a pre fetch that the UI uses so this isn't required.

PR #1708, opened 2022-07-06
Remove unneeded dependencies

StuAA78:
https://github.com/DEFRA/water-abstraction-team/issues/26

Our `package.json` file has a number of dependencies hanging around which are no longer used. We identify them using `npx depcheck` and uninstall them accordingly.

Note that the following dependencies were identified as not being used, but they are used as far as we're aware, just not referenced in a way that depcheck recognises:
- `bookshelf-case-converter-plugin`
- `bookshelf-json-columns`
- `db-migrate-pg`

We make a couple of small code changes:
- One file requires `@hapi/joi` but in `package.json` we are including `joi` -- `@hapi/joi` only works as it's currently introduced as a dependency of a dependency. We therefore update the file to directly use `joi`.
- One file requires `request-promise` which again is only being introduced as a dependency of a dependency. Elsewhere in the code we use `request-promise-native`, which is included in `package.json`, so we update this file to follow suit.

Finally, we install some packages which are only currently included as dependencies of dependencies, to ensure that we are explicit about which packages our service uses.


Jozzey (approved):
👍🏼

PR #1766, opened 2022-08-10
Set up background tasks module

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3745

We are separating out our background tasks. We start by moving the creation of workers from QueueManager to the background service. QueueManager sends a request to the bg service so to anything that calls QueueManager to create a queue, there is no outward change.


Cruikshanks:
Closed in favour of https://github.com/DEFRA/water-abstraction-service/pull/1784

PR #1787, opened 2022-08-23
Add `isBackground` flag to `config.js`

StuAA78:
We have a few places where we check whether or not we're running in the background service. We therefore centralise this into `config.js`


Jozzey (approved):
👍🏼

Cruikshanks (approved):

PR #1794, opened 2022-08-25
Remove catbox-redis from service

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/46

Catbox-redis is only used in the service to cache the address search data. The reason given for it's use is "caching avoids API rate limit being exceeded on subsequent identical requests".

As we are using v1.1 of the EA Address Facade which was developed with rate limiting built in this cache is not required and can be removed. As this is the only place catbox-redis is used the package can be removed once the cache has been removed from the address lookup plugin.


test/modules/address-search/controller.test.js
```diff
@@ -23,25 +25,22 @@ experiment('modules/application-state/controller', () => {
...
       result = await controller.getAddressSearch(request)
     })
 
     test('fetches the data via the server method', async () => {
-      expect(request.server.methods.getAddressesByPostcode.calledWith('TT1 1TT')).to.be.true()
+      expect(addressService.getAddressesByPostcode.calledWith('TT1 1TT')).to.be.true()
     })
 
     test('returns the data', async () => {
+      console.log(result)
```
Cruikshanks, line 43: Guessing this was for debugging? Think we can remove it now 😁

test/modules/address-search/controller.test.js
```diff
@@ -23,25 +25,22 @@ experiment('modules/application-state/controller', () => {
...
       }
+      sandbox.stub(addressService, 'getAddressesByPostcode').resolves({
+        data: [
+          new Address()
+        ]
+      })
       result = await controller.getAddressSearch(request)
     })
 
     test('fetches the data via the server method', async () => {
```
Cruikshanks, line 38: I'd argue this whole test can 'do-one'. We shouldn't be testing the internals of a method i.e. did the public method call some private method IMHO.

Cruikshanks (changes requested):
A `console.log()` definitely needs to be deleted. But I've also suggested that the test you've had to alter can go completely. What do you think?

test/modules/address-search/controller.test.js
Jozzey: Yes, I'll get shot of that

Cruikshanks (approved):

PR #1803, opened 2022-09-02
Use new `url-join` helper

StuAA78:
https://github.com/DEFRA/water-abstraction-helpers/pull/183

To cut down our reliance on external dependencies where possible, we are rewriting the `url-join` dependency to use more standard JS url handling instead of custom regexes. This PR replaces all uses of `url-join` with the one in the new helper.


Cruikshanks:
We are no longer investing time and effort into improving the legacy codebases. Instead, we are focusing on migrating functionality to [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system).

PR #1810, opened 2022-09-08
Update `libxmljs`

StuAA78:
We have an outstanding dependabot PR to update `libxmljs`. However the unit tests fail, so we need to make some changes in order to safely update.


StuAA78:
Resolved in https://github.com/DEFRA/water-abstraction-service/pull/1812

PR #1812, opened 2022-09-08
Fix libxmljs update issue

Jozzey:
The latest version of libxmljs 0.19.10 does not put the binary in the correct location when installing locally.

A postinstall script has been added to check if the binary is in the correct location and to move it if it isn't.


Cruikshanks (changes requested):
Thanks for sorting this chaps. Just a request for some minor tweaks if that's ok

- can we rename the file to use kebab case and match our general convention, for example `fix-libxmljs.sh`
- can we make the file executable to allow us to run it directly. To do so run `chmod +x fix-libxmljs.sh`

Those I'd argue should be done. The following are subjective

- we have a `scripts/` folder. Would it be better to locate the file there instead of in the root?
- take the file rename further and make it more descriptive, for example `fix-libxmljs-location.sh`

Jozzey:
> 

👍🏼 I'll make those tweaks and push it back up

Cruikshanks (approved):

PR #1824, opened 2022-09-21
Rewrite presroc supplementary bill run process

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3719

As part of the process of understanding how the presroc supplementary billing process works, we are rewriting the jobs used in the process to gain greater understanding of it.

Note that we may not be merging this once completed.


Cruikshanks:
Going to close this now. We did [re-write supplementary billing](https://github.com/DEFRA/water-abstraction-system). But in 💯 our own way using code that made far more sense.

We no longer feel the need to refer to this anymore so closing to remove the noise/nag in our PR's and branches.

PR #1841, opened 2022-10-13
Swap out or remove Deprecated packages

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/22

We are finding as we go through the code base that we have a number of deprecated packages.

They are no longer maintained so we need to get them out. If we don't, we expose ourselves to unfixed errors, vulnerabilities, and issues when we come to upgrade other packages.

The packages `@sindresorhus/is`, `bullmq`, `csv-parse`, `got`, `knex`, `notifications-node-client`, `p-waterfall` & `url-join` have not been updated as part of this PR as they are either EMCAScript only modules or significantly break the service when updated. These will be done in separate PRs.


Jozzey:
Updated version of knex depends on updating the version of bullmq so would probably be best to update both of these at the same time. Updating these packages will break the app so I have left them for a day other than late Friday afternoon 😄

Jozzey:
Can't bump major version of `p-waterfall` package as the new version is ESM only

Cruikshanks (approved):

PR #1868, opened 2022-10-27
Add new SROC supplementary billing test data

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3787

In order to test our new code to identify licences that need to be included in an SROC supplementary bill run we need some test data.

We want to ensure we have this in place rather than having to manual edit whatever licences happen to be at hand at the time we start to code.

So, this change adds a new test base for our SROC supplementary billing work.


Jozzey (approved):
👍🏼

PR #1885, opened 2022-11-14
Stop passing err to logger

Beckyrose200:
https://github.com/DEFRA/water-abstraction-team/issues/45

Here we are trying to make the error logs more usable. Currently when an error happens the logs are getting jammed up with thousands of lines making them difficult to read and unusable. This change stops that pollution making them more usable.

### DIFF
--- a/index-background.js
+++ b/index-background.js
@@ -86,12 +86,12 @@ const start = async function () {
       server.log('info', `Service ${name} running at: ${uri}`)
     }
   } catch (err) {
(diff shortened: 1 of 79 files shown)


Cruikshanks (approved):
You've unfortunately been penalised because you've touched code files that already have poor test coverage.

If we had all the time in the world we'd get it bumped up. But we don't. So, happy to approve and ignore SonarCloud on this one.

raspberry3

PR #1893, opened 2022-11-21
Trigger SROC supplementary from PRESROC

StuAA78:
https://eaflood.atlassian.net/browse/WATER-3828

When PRESROC supplementary billing has completed, we want to notify the `water-abstraction-system` to trigger SROC supplementary billing.


StuAA78:
Following a review of our approach we've decided that we'd be better off having the UI trigger the sroc supplementary bill run, instead of hooking into the flow of jobs in the service.

PR #1962, opened 2023-01-19
Removing Lodash 'Pick' function

Beckyrose200:

This PR is focusing on the 'pick' function

### DIFF
--- a/integration-tests/billing/services/fixture-loader/adapters/BookshelfAdapter.js
+++ b/integration-tests/billing/services/fixture-loader/adapters/BookshelfAdapter.js
@@ -1,7 +1,5 @@
 'use strict'
 
-const { pick } = require('lodash')
-
 class BookshelfAdapter {
   constructor (bookshelf) {
     this._bookshelf = bookshelf
@@ -34,7 +32,7 @@ class BookshelfAdapter {
       if (err.code === '23505' && (err.constraint in constraints) && ref) {
         const model = await this._bookshelf
           .model(modelName)
-          .forge(pick(data, constraints[err.constraint]))
+          .forge(this._pick(data, constraints[err.constraint]))
           .fetch()
 
         return model.toJSON()
@@ -42,6 +40,16 @@ class BookshelfAdapter {
         throw err
       }
     }
+  }
+
+  _pick (obj, props) {
+    const picked = {}
+    for (const prop of props) {
+      if (obj[prop] !== undefined) {
+        picked[prop] = obj[prop]
+      }
+    }
+    return picked
   }
 }
 
--- a/src/modules/billing/jobs/sync-charge-categories.js
+++ b/src/modules/billing/jobs/sync-charge-categories.js
@@ -2,7 +2,6 @@
 'use strict'
 
 // Dependencies
-const { isEqual } = require('lodash')
 const moment = require('moment')
 const BlueBirdPromise = require('bluebird')
 const csvParse = BlueBirdPromise.promisify(require('csv-parse'))
@@ -32,14 +31,6 @@ const createMessage = () => ([
   }
 ])
 
-const pick = (obj, props) => {
-  const picked = {}
-  for (const prop of props) {
-    picked[prop] = obj[prop]
-  }
-  return picked
-}
-
 const handler = async () => {
   logger.info(`${JOB_NAME}: Job has started`)
 
@@ -59,20 +50,8 @@ const handler = async () => {
     const mappedChargeCategory = chargeCategoriesMapper.csvToModel(arraysFromCSV[i])
     const chargeCategoryExists = await chargeCategoriesRepo.findOneByReference(mappedChargeCategory.reference)
 
-    const keys = [
-      'description',
-      'shortDescription',
-      'subsistenceCharge',
-      'minVolume',
-      'maxVolume',
-      'isTidal',
-      'lossFactor',
-      'modelTier',
-      'isRestrictedSource'
-    ]
-
     if (chargeCategoryExists) {
-      if (!isEqual(pick(chargeCategoryExists, keys), pick(mappedChargeCategory, keys))) {
+      if (_updateExistingCategory(chargeCategoryExists, mappedChargeCategory)) {
         logger.info(`Updating an existing charge category: ${mappedChargeCategory.reference}`)
         await chargeCategoriesRepo.updateByReference(mappedChargeCategory.reference, mappedChargeCategory)
       }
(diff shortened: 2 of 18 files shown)


src/modules/billing/services/supplementary-billing-service/data-service.js
```diff
@@ -46,6 +46,14 @@ const pickKeys = [
   'isNewLicence'
 ]
 
+const pick = (obj, props) => {
```
Beckyrose200, line 49: I chose to add a function here for pick otherwise I would have had to duplicate the pickKeys which would have been an additional 25 lines of code

src/modules/billing/jobs/sync-charge-categories.js
```diff
@@ -32,6 +32,14 @@ const createMessage = () => ([
   }
 ])
 
+const pick = (obj, props) => {
```
Beckyrose200, line 35: I chose to add a function here for pick otherwise I would have had to duplicate the keys which would have been an additional 18 lines of code (since pick is used twice in this file)

src/modules/billing/services/supplementary-billing-service/supplementary-processor.js
```diff
@@ -47,6 +47,14 @@ const commonTransactionKeys = [
   'financialYearEnding'
 ]
 
+const pick = (obj, props) => {
```
Beckyrose200, line 50: I chose to add a function here for pick otherwise I would have had to duplicate the commonTransactionKeys and abstractionPeriodKeys which would have been an additional 24 lines of code (since pick is used three times in this file)

test/modules/billing/services/supplementary-billing-service/data-service.test.js
```diff
@@ -26,6 +25,14 @@ const batchId = uuid()
 
 const { createTransaction } = require('./test-helpers')
 
+const pick = (obj, props) => {
```
Beckyrose200, line 28: I chose to add a function here for pick otherwise I would have had to duplicate the keys which would have been an additional 50 lines of code (since pick is used twice in this file)

integration-tests/billing/services/fixture-loader/adapters/BookshelfAdapter.js
```diff
@@ -43,6 +41,16 @@ class BookshelfAdapter {
       }
     }
   }
+
```
Beckyrose200, line 44: This integration test file isn't being run and so I added the pick function as I know this is working. This file will soon be deleted as part of an issues ticket the team is working on.

src/modules/billing/jobs/sync-charge-categories.js
Cruikshanks: Don't worry about adding lines of code if it makes things easier to read and understand. In this case, a new team member would have to work through `pick()` to understand what it is doing. IMHO seeing two objects created from a subset of others is understandable at a glance, even if the keys are repeated.

However, in this case, I believe there is a 3rd way. Having looked at the context where `pick()` is being used I think we can refactor and make the intent more explicit.


```javascript
const handler = async () => {
  logger.info(`${JOB_NAME}: Job has started`)

  // Get the Body and the ETag of the S3 object
  const { Body, ETag } = await s3Connector.getObject(csvKey)

  // Get the current application state for the import
  const { data: currentApplicationState } = await applicationState.get('charge-categories-import')

  if (ETag === currentApplicationState.etag) {
    logger.info('No change detected. Not processing file.')
    return 'No change detected. Not processing file.'
  }

  const arraysFromCSV = await csvParse(Body, { columns: true })
  for (let i = 0; i < arraysFromCSV.length; i++) {
    const mappedChargeCategory = chargeCategoriesMapper.csvToModel(arraysFromCSV[i])
    const chargeCategoryExists = await chargeCategoriesRepo.findOneByReference(mappedChargeCategory.reference)

    if (chargeCategoryExists) {
      if (_updateExistingCategory(chargeCategoryExists, mappedChargeCategory)) {
        logger.info(`Updating an existing charge category: ${mappedChargeCategory.reference}`)
        await chargeCategoriesRepo.updateByReference(mappedChargeCategory.reference, mappedChargeCategory)
      }
    } else {
      logger.info(`Creating a new charge category: ${mappedChargeCategory.reference}`)
      await chargeCategoriesRepo.create(mappedChargeCategory)
    }
  }

  return applicationState.save('charge-categories-import', { etag: ETag })
}

const _updateExistingCategory = (existingCategory, mappedCategory) => {
  const keys = [
    'description',
    'shortDescription',
    'subsistenceCharge',
    'minVolume',
    'maxVolume',
    'isTidal',
    'lossFactor',
    'modelTier',
    'isRestrictedSource'
  ]

  for (const key of keys) {
    if (existingCategory[key] !== mappedCategory[key]) {
      return true
    }
  }

  return false
}
```

`pick()` is used to extract the values to compare to determine whether an existing charge category needs to be updated. Knowing that we can refactor things so that

- we don't depend on `lodash.pick()`
- we don't depend on `lodash.isEqual()`
- we only declare our `keys` once
- we can create a function with one purpose, and name it based on that purpose

A new member still has to look at `_updateExistingCategory()` to see what it's doing. But hopefully, the name of our function gives them a big clue. Compare that to

```javascript
const pick = (obj, props) => {
  const picked = {}
  for (const prop of props) {
    picked[prop] = obj[prop]
  }
  return picked
}

// ...

if (!isEqual(pick(chargeCategoryExists, keys), pick(mappedChargeCategory, keys))) {
 // ...
}
```

which is doing the same thing but the functionality is spread throughout the module and the meat of the logic is packed inline in the `if()`. 🤷

src/modules/billing/services/supplementary-billing-service/data-service.js
Cruikshanks: I think you should be okay here. I can only see one reference to `pick()` and `pickKeys` which should mean you wouldn't need to repeat `pickKeys`. Happy to take a look with you!

src/modules/billing/services/supplementary-billing-service/supplementary-processor.js
Cruikshanks: Again, I think the way those key arrays are being used means we can use them without needing to repeat them and avoid having to implement a `pick()` method within the module.

test/modules/billing/services/supplementary-billing-service/data-service.test.js
Cruikshanks: In the test where `pick()` is used, if we refactor how it is doing the assertion we can remove the need to have a `pick()` function.

```javascript
// Do this
for (const key of matchingKeys) {
  expect(transactions[0][key]).to.equal(createdTransaction[key])
}

// instead of this
expect(pick(transactions[0], matchingKeys)).to.equal(pick(createdTransaction, matchingKeys))
```

Cruikshanks (changes requested):
Because I benefit from having being doing this stuff too long(!) I'm able to see the ways I think we can cast `pick()` out of the code altogether.

Shout if you need to get together to understand my suggestions or if you don't think they help.

PR #1969, opened 2023-01-25
Remove Lodash Get Function

Beckyrose200:

This PR is focusing on the 'get' function

### DIFF
--- a/src/modules/users/controller.js
+++ b/src/modules/users/controller.js
@@ -1,6 +1,8 @@
+'use strict'
+
 const idmConnector = require('../../lib/connectors/idm')
 const Boom = require('@hapi/boom')
-const { get, partial } = require('lodash')
+const { partial } = require('lodash')
 const crmEntitiesConnector = require('../../lib/connectors/crm/entities')
 const crmDocumentsConnector = require('../../lib/connectors/crm/documents')
 const idmUserRolesConnector = require('../../lib/connectors/idm/user-roles')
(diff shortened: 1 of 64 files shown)


src/modules/users/controller.js
```diff
@@ -218,8 +220,8 @@ const getStatus = async (request, h) => {
     data: {
```
Cruikshanks, line 220: I can't get GitHub to highlight or put my suggestion in the right place for this one. But I would suggest refactoring this. First, we can combine extracting and parsing the data. Then I would make the return statement simpler by calling the functions first and using their return values.

```javascript
  const results = await getUserCompanyStatus(userResponse.data)

  const companiesData = results[0].data?.companies ?? []
  const verificationsData = results[1].data ?? []
  const documentHeaders = results[2] ?? []
  const licencesMap = results[3]

  const user = mapUserStatus(userResponse.data)
  const companies = mapCompanies(
    companiesData,
    verificationsData,
    documentHeaders,
    licencesMap
  )

  return {
    data: {
      user,
      companies
    },
    error: null
  }
```

Cruikshanks (changes requested):
The vast bulk of these suggestions is replacing where we are using ternary operators to check something exists else use an alternative.

In these instances, we can make use of the JavaScript [Nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).

TBH be honest its something @StuAA78 has only recently introduced me to. But I think in these cases it makes the code cleaner and more readable.

Within all those suggestions, I've also got a couple of others for how I think we can make things clearer, by splitting single lines of logic into multiple steps.

Shout if you need me to go over anything I've suggested here.

Cruikshanks (approved):

Cruikshanks:
Ignore the SonarCloud failure. I can see now it's just code coverage and that's not your responsibility. You just happen to have touched the files!

PR #1970, opened 2023-01-25
Add default timeout to Got requests

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-3883

When working on WATER-3833 we were trying to determine what the legacy code does in the event of an error when running a bill run. The quickest way to achieve that locally was to turn off the [Charging Module API](https://github.com/DEFRA/sroc-charging-module-api) so all requests to it would fail.

What we found is they don't fail, at least not locally. Instead, the service sits there in-perpetuity waiting for a response. The only reason they eventually fail in our AWS environments is that the requests go through proxy servers that will eventually timeout.

After checking, the reason is that the previous team didn't set a default timeout for these requests. The [Got](https://github.com/sindresorhus/got) package states in its documentation

> It is a good practice to set a timeout to prevent hanging requests.
> By default, there is no timeout set.

_[Timeout options](https://github.com/sindresorhus/got/blob/main/documentation/6-timeout.md)_

The legacy code uses a number of packages for making requests (😩). We're only interested in the ones with the charging module which all happen to use **Got**. To make the process more robust and reduce the chance of calls to us during supplementary bill run testing, we're adding a default timeout of 5 seconds (this is what we use in our new code).


Jozzey (approved):
👍🏼

PR #1998, opened 2023-02-09
Remove 'isObject', 'first', 'isUndefined', and 'find' lodash function

Beckyrose200:

This PR is focusing on the 'isObject', 'first', 'isUndefined', and 'find' functions

### DIFF
--- a/src/lib/contact-list.js
+++ b/src/lib/contact-list.js
@@ -27,7 +27,7 @@ function getPreferredContact (contacts, rolePriority) {
       return acc
     }
 
-    return contacts.find(o => o.role === role)
+    return contacts.find((o) => o.role === role)
   }, null)
 }
 
@@ -41,7 +41,7 @@ function createSendList (licences, rolePriority) {
 
   licences.forEach(licence => {
     // Get relevant contacts
-    const licenceHolder = licence.contacts.find(o => o.role === 'licence_holder')
+    const licenceHolder = licence.contacts.find((o) => o.role === 'licence_holder')
 
     // Get preferred notification contact
     // In future this may need to support sending specific messages to differnet
--- a/src/lib/contact-list.js
+++ b/src/lib/contact-list.js
@@ -1,3 +1,5 @@
+'use strict'
+
 /**
  * Gets a filtered list of licences together with all
  * attached contacts, from the CRM.
@@ -10,7 +12,6 @@
  */
 
 const Boom = require('@hapi/boom')
-const { find } = require('lodash')
 const sha1 = require('sha1')
 const { getDocumentContacts } = require('./connectors/crm/documents')
 
@@ -25,7 +26,8 @@ function getPreferredContact (contacts, rolePriority) {
     if (acc) {
       return acc
     }
-    return find(contacts, { role })
+
+    return contacts.find(o => o.role === role)
   }, null)
 }
 
@@ -39,7 +41,7 @@ function createSendList (licences, rolePriority) {
 
   licences.forEach(licence => {
     // Get relevant contacts
-    const licenceHolder = find(licence.contacts, { role: 'licence_holder' })
+    const licenceHolder = licence.contacts.find(o => o.role === 'licence_holder')
 
     // Get preferred notification contact
     // In future this may need to support sending specific messages to differnet
--- a/src/lib/licence-transformer/nald-transformer.js
+++ b/src/lib/licence-transformer/nald-transformer.js
@@ -4,7 +4,7 @@
  * Transforms NALD data into VML native format
  * @module lib/licence-transformer/nald-transformer
  */
-const { find, uniqBy } = require('lodash')
+const { uniqBy } = require('lodash')
 const { sentenceCase } = require('sentence-case')
 
 const BaseTransformer = require('./base-transformer')
@@ -27,7 +27,7 @@ class NALDTransformer extends BaseTransformer {
 
     const mostRecentVersionPurposes = data.data.purposes.filter(purpose => purpose.AABV_INCR_NO !== data.data.versions.sort((a, b) => parseInt(b.INCR_NO) - parseInt(a.INCR_NO)))
 
-    const licenceHolderParty = find(currentVersion.parties, (party) => {
+    const licenceHolderParty = currentVersion.parties.find((party) => {
       return party.ID === currentVersion.ACON_APAR_ID
     })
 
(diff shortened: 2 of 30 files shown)


src/lib/licence-transformer/nald-transformer.js
```diff
@@ -27,7 +27,7 @@ class NALDTransformer extends BaseTransformer {
 
     const mostRecentVersionPurposes = data.data.purposes.filter(purpose => purpose.AABV_INCR_NO !== data.data.versions.sort((a, b) => parseInt(b.INCR_NO) - parseInt(a.INCR_NO)))
 
-    const licenceHolderParty = find(currentVersion.parties, (party) => {
+    const licenceHolderParty = currentVersion.parties.find((party) => {
       return party.ID === currentVersion.ACON_APAR_ID
     })
```
StuAA78, line 32: If we're okay with one-line syntax for `find` (as we've used elsewhere) then this and other similar ones could be one line:
```suggestion
    const licenceHolderParty = currentVersion.parties.find((party) => party.ID === currentVersion.ACON_APAR_ID)
```

src/modules/batch-notifications/config/returns/lib/create-notification-data.js
```diff
@@ -99,7 +98,7 @@ const templateRandomiser = templateType => {
 
 const getRelevantRowData = (rows, reminderRef) => {
   const contactAndMessageType = reminderRef.substring(17) // remove 'returns_invitation' from beginning
-  return first(rows.filter(row => row.message_ref.includes(contactAndMessageType)))
+  return (rows.filter(row => row.message_ref.includes(contactAndMessageType)))[0]
```
StuAA78, line 101: We've split this kind of thing across two lines elsewhere, it might be clearer to do it here as well? eg:
```suggestion
  const filteredRows = rows.filter((row) => row.message_ref.includes(contactAndMessageType))
  return filteredRows[0]
```

src/lib/contact-list.js
```diff
@@ -25,7 +26,8 @@ function getPreferredContact (contacts, rolePriority) {
     if (acc) {
       return acc
     }
-    return find(contacts, { role })
+
+    return contacts.find(o => o.role === role)
```
StuAA78, line 30: I think we usually put brackets around single parameters so this and others would need to be changed.
```suggestion
    return contacts.find((o) => o.role === role)
```

StuAA78 (changes requested):
Looks great! Just maybe needs a few tweaks for consistency.

src/modules/batch-notifications/lib/batch-notifications.js
```diff
@@ -17,7 +16,7 @@ const loadJobData = async (eventId) => {
   }
 
   // Load config
-  const config = find(configs, { messageType: event.subtype })
+  const config = configs.find(o => o.messageType === event.subtype)
```
StuAA78, line 19:
```suggestion
  const config = configs.find((o) => o.messageType === event.subtype)
```

src/modules/returns/lib/returns-upload-validator.js
```diff
@@ -152,26 +152,31 @@ const validateLineFrequency = ret => {
  * @param  {Object} context - additional context data
  * @return {Boolean}         true if document was found
  */
-const validateLicence = (ret, context) =>
-  find(context.documents, getDocumentFilter(ret.licenceNumber))
+const validateLicence = (ret, context) => {
+  const externalId = getDocumentFilter(ret.licenceNumber)
+  return context.documents.find(o => o.system_external_id === externalId.system_external_id)
```
StuAA78, line 157:
```suggestion
  return context.documents.find((o) => o.system_external_id === externalId.system_external_id)
```

src/modules/returns/lib/returns-upload-validator.js
```diff
@@ -152,26 +152,31 @@ const validateLineFrequency = ret => {
...
  * Checks that the user's company matches that of the CRM document
  * @param  {Object} ret     - uploaded return
  * @param  {Object} context - additional context data
  * @return {Boolean}         true if company matches
  */
-const validatePermission = (ret, context) =>
-  find(context.documents, getCompanyDocumentFilter(ret.licenceNumber, context.companyId))
+const validatePermission = (ret, context) => {
+  const result = getCompanyDocumentFilter(ret.licenceNumber, context.companyId)
+  return context.documents.find(o => o.system_external_id === result.system_external_id && o.company_entity_id === result.company_entity_id)
```
StuAA78, line 168:
```suggestion
  return context.documents.find((o) => o.system_external_id === result.system_external_id && o.company_entity_id === result.company_entity_id)
```

src/modules/returns/lib/returns-upload-validator.js
```diff
@@ -152,26 +152,31 @@ const validateLineFrequency = ret => {
...
 /**
  * Checks a return was found in the returns service for the uploaded return ID
  * @param  {Object} ret     - uploaded return
  * @param  {Object} context - additional context data
  * @return {Boolean}         true if return found
  */
-const validateReturnExists = (ret, context) =>
-  find(context.returns, { return_id: ret.returnId })
+const validateReturnExists = (ret, context) => {
+  return context.returns.find(o => o.return_id === ret.returnId)
```
StuAA78, line 178:
```suggestion
  return context.returns.find((o) => o.return_id === ret.returnId)
```

src/lib/models/contact-list.js
```diff
@@ -24,7 +24,7 @@ class ContactList {
    * @return {Contact}      - the contact (if found)
    */
   getByRole (role) {
-    return find(this.contacts, contact => compareRoles(contact.role, role))
+    return this.contacts.find(contact => compareRoles(contact.role, role))
```
StuAA78, line 27:
```suggestion
    return this.contacts.find((contact) => compareRoles(contact.role, role))
```

src/lib/object-mapper.js
```diff
@@ -90,7 +89,7 @@ class Mapper {
       }
 
       // Undefined values in the source are skipped
-      if (values.every(isUndefined)) {
+      if (values.every(val => val === undefined)) {
```
StuAA78, line 92:
```suggestion
      if (values.every((val) => val === undefined)) {
```

StuAA78 (changes requested):
Just a handful of single parameters without brackets that slipped through the net, otherwise all looks good to me 👍

StuAA78 (approved):

PR #2050, opened 2023-03-13
Remove lodash code

Beckyrose200:


Cruikshanks (changes requested):
We've got this exact same function and comment in 4 places

```javascript
const { reduce, isPlainObject } = require('lodash')

/* This regex is converting any string into Camel Case
 * [^a-zA-Z0-9] is matching any character except those inside the square brackets.
 * This could be a dash (-), space ( ), or underscore (_)
 * +(.) is matching the first character after the dash, space or underscore.
 * /g. Replaces all the global matches and not just the first match, so the whole string can be converted
*/
const toCamelCase = (key) => {
  return key.replace(/[^a-zA-Z0-9]+(.)/g, (match, char) => char.toUpperCase())
}
```

That certainly meets the criteria for [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))! So, we need to move this into a single place, and `require()` that module for use in these files.

There might be an existing module (looking at you `src/lib/camel-case-keys.js`) you can add the function to or a completely new one might be more appropriate. I leave that to you to decide. 😁

PR #2056, opened 2023-03-15
Remove constraint from `billing_transactions` table

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3906

Somehow (seriously, how!? :angry: :sob: ) our local DB schema does not match the `dev` schema (and most likely all AWS DB schemas).

Our solution is premised on the constraint `CONSTRAINT billing_transactions_billing_invoice_licence_id_fkey FOREIGN KEY (billing_invoice_licence_id) REFERENCES water.billing_invoice_licences(billing_invoice_licence_id)` not being there. So, we just need to throw together a fix migration and re-deploy to `dev` and all will be well. :weary:

### DIFF
--- a/migrations/20230315122011-remove-constraint-from-billing-transactions.js
+++ b/migrations/20230315122011-remove-constraint-from-billing-transactions.js
@@ -1,0 +1,47 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+let Promise
+
+/**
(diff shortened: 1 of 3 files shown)


Jozzey:
I deployed this branch to DEV to check if the migration works OK and all looks good.

Cruikshanks (approved):

PR #2059, opened 2023-03-16
Remove 'isFunction' and 'identity' lodash functions

Beckyrose200:

Lodash is used a lot in our codebase. Although we are not totally removing this module, some of the functions in the code using lodash are unnecessary. Here we are removing the unnecessary ones are replacing it with vanilla js. This PR focuses on removing 'isFunction' and 'identity'.


src/modules/billing/mappers/transaction.js
```diff
@@ -241,7 +242,7 @@ const mapChargeModuleDate = str =>
 const mapAgreementsToChargeModule = transaction => {
   const section130Agreement = ['S130U', 'S130S', 'S130T', 'S130W']
```
StuAA78, line 243:
```suggestion
  // Sets a boolean based on whether at least one agreement with a section 130 code could be found
  const section130Agreement = ['S130U', 'S130S', 'S130T', 'S130W']
```

StuAA78 (changes requested):
Couple of small comment tweaks but otherwise all good! 😁

Cruikshanks (approved):

PR #2061, opened 2023-03-17
Remove 'uniq' lodash function

Beckyrose200:

Lodash is used a lot in our codebase. Although we are not totally removing this module, some of the functions in the code using lodash are unnecessary. Here we are removing the unnecessary ones are replacing it with vanilla js. This PR focuses on removing 'uniq'.


StuAA78 (changes requested):
Looks good! The only change I'd request would be adding comments (or amending existing ones) to explain that we're using Set to remove duplicate values.

StuAA78 (approved):

PR #2062, opened 2023-03-17
Fix error with internal search for users

Cruikshanks:
Just spotted in regression testing that we have a bug with the internal search. If you enter a user email that exists the search errors causing a `500` to be returned to the client. In our AWS environments that results in a WAF violation and the user seeing the WAF error page.

The problem is caused by not all users having scopes attached to their roles. A quick tweak and all is well.


Jozzey (approved):
👍🏼

StuAA78 (approved):

PR #2065, opened 2023-03-20
Remove 'chunk' lodash function

Beckyrose200:

Lodash is used a lot in our codebase. Although we are not totally removing this module, some of the functions in the code using lodash are unnecessary. Here we are removing the unnecessary ones are replacing it with vanilla js. This PR focuses on removing 'chunk'.


src/lib/chunk.js
```diff
@@ -0,0 +1,10 @@
+'use strict'
+
+const chunk = (arr, chunkSize = 1, cache = []) => {
+  const tmp = [...arr]
+  if (chunkSize <= 0) return cache
+  while (tmp.length) cache.push(tmp.splice(0, chunkSize))
+  return cache
+}
+
+module.exports = chunk
```
Cruikshanks, line 10: I should have given a bit more guidance when it comes to the Lodash work. We're not trying to do a straight swap for what Lodash is doing. Instead, where we see it used we are trying to identify

- is it needed?
- could the logic be tackled in a different way?
- if there is a need, what solution would be specific for our service?

`chunk()` to me seems to fall into the latter. We make the call in so many places, and we are using it in a very 'vanilla' way. So, I agree, it doesn't make sense to repeat the chunking code everywhere we do it.

So, the first thing I should have made clearer is that when we do add a whole new module/service it should follow our conventions. If you need examples check out anything in [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system). If we are amending existing code, we follow what we find though we still avoid not using brackets, inline blocks etc.

On the implementation itself, having looked through all the places we are using `chunk()` we are only passing through an array and a chunk size. This means we should tailor the Lodash replacement code to our use case, rather than implementing a generic solution as Lodash provides.

```suggestion
'use strict'

/**
 * Use to break arrays up into 'chunks'
 * @module Chunk
 */

/**
 * Returns an array of the array to chunk 'chunked' into smaller arrays
 *
 * Often needed where you have something you need to break down into batches, for example, records to persist to a
 * DB. You pass in the array of records along with your batch size, and this will return an array of smaller arrays
 * of the size you specified.
 * 
 * When the array is not fully divisible by the `chunkSize` any remainders will be returned in the last array chunk. If
 * the `chunkSize` is greater than the array size you'll just an array with an array inside!
 * 
 * If `chunkSize` is 0 or less you'll get an empty array returned
 * 
 * @param {Array} arrayToChunk The array to break into smaller chunks
 * @param {Number} chunkSize What size 'chunks' the array should be broken up into. Note, the last chunk may be equal or
 * less than `chunkSize`
 * 
 * @returns {Array[Array]} a single array containing the `arrayToChunk` split into smaller arrays
 */
function chunk (arrayToChunk, chunkSize) {
  const tmp = [...arrayToChunk]
  const cache = []
  
  if (chunkSize <= 0) {
    return cache
  }
  
  while (tmp.length) {
    cache.push(tmp.splice(0, chunkSize))
  }
  
  return cache
}

module.exports = {
  chunk
}
```

Cruikshanks (changes requested):

PR #2068, opened 2023-03-21
Refactor removal of lodash code

Beckyrose200:

In our code base, we've been gradually replacing Lodash code with vanilla JavaScript to reduce the library's usage. However, we've found a few Lodash functions that we still need to use in certain parts of the code. To better organize these functions, we created separate modules for them, which were then called throughout the code base. After creating several of these modules, we decided it would be cleaner to move all of the Lodash modules into a single file and export them from there. This pull request implements that change.


src/lib/object-helpers.js
```diff
@@ -0,0 +1,88 @@
+'use strict'
+
+/**
+ * Use to break arrays up into 'chunks'
+ * @module Chunk
+ */
```
Cruikshanks, line 6: Just a copy & paste thing to correct! 😁 

```suggestion
/**
 * Collection of helper functions that allowed us to move away from Lodash
 * @module ObjectHelpers
 */
```

Cruikshanks (changes requested):
It's so tidy now! 😍 

Just a quick correction in the module documentation.

StuAA78 (approved):

Cruikshanks (approved):

PR #2073, opened 2023-03-23
Remove 'reduce' and 'sortBy' lodash function

Beckyrose200:

Lodash is used a lot in our codebase. Although we are not totally removing this module, some of the functions in the code using lodash are unnecessary. Here we are removing the unnecessary ones are replacing it with vanilla js. This PR focuses on removing 'reduce' and 'sortBy'.


src/modules/charge-versions-upload/lib/validator/index.js
```diff
@@ -114,12 +116,32 @@ const validateGroups = async (rows, headings, jobName) => {
...
+    } else {
+      if (a.chargeReferenceDetailsChargeElementGroup < b.chargeReferenceDetailsChargeElementGroup) {
+        return -1
+      } else if (a.chargeReferenceDetailsChargeElementGroup > b.chargeReferenceDetailsChargeElementGroup) {
+        return 1
+      } else {
+        return 0
+      }
+    }
+  })
```
Cruikshanks, line 143: In `src/modules/charge-versions-upload/lib/mapper/index.js` you opted to do the same thing by creating a new `sortBy()` method and using ternary `if` statements to reduce the nesting.

Of the 2 approaches I think it was clearer. So I would suggest seeing if there is a way to refactor this change to use the same approach. What do you think? 🤷

Cruikshanks (changes requested):
This is all I've got for this change. Other than that it looks 🍑y! 😁

src/lib/services/charge-versions.js
```diff
@@ -117,10 +116,12 @@ const getStartDate = chargeVersion => chargeVersion.dateRange.startDate
  * @param {Array<ChargeVersion>} chargeVersions
  */
 const refreshEndDates = chargeVersions => {
-  const filtered = chargeVersions.filter(isCurrentChargeVersion)
-  const sorted = sortBy(filtered, getStartDate)
+  const filteredAndSorted = chargeVersions.filter(isCurrentChargeVersion)
+  filteredAndSorted.sort((a, b) => {
+    return new Date(getStartDate(a)) - new Date(getStartDate(b))
+  })
```
StuAA78, line 122: It might be a little nicer to chain `.sort()` instead of assigning the filtered value to then mutating it, and we probably want to be explicit about what we're passing in the `.filter()`):
```suggestion
  const filteredAndSorted = chargeVersions
    .filter((chargeVersion) => {
      return isCurrentChargeVersion(chargeVersion)
    })
    .sort((a, b) => {
      return new Date(getStartDate(a)) - new Date(getStartDate(b))
    })
```

src/modules/billing/services/charge-processor-service/lib/agreements.js
```diff
@@ -50,10 +50,10 @@ const getAgreementsHistory =
 (chargePeriod, licenceAgreements) => {
   // Filter out agreements that are not S127/S130 as they don't affect charging
   // And sort by start date
-  const filteredAndSorted = sortBy(
-    licenceAgreements.filter(isBillingAgreement),
-    getLicenceAgreementStartDate
-  )
+
+  const filteredAndSorted = licenceAgreements
+    .filter(isBillingAgreement)
```
StuAA78, line 55: See earlier comment about being explicit about what we pass to the `.filter()`

src/modules/billing/services/supplementary-billing-service/supplementary-processor.js
```diff
@@ -203,7 +203,9 @@ const getPairGroupingKey = transaction => {
  * @return {Array<Object>}
  */
 const filterCancellingTransactions = transactions => {
```
StuAA78, line 205: As before we can chain this instead of mutating, it won't let me make a suggestion for the whole code but if we wanted we could have something like this:
```js
const filterCancellingTransactions = transactions => {
  return transactions
    .sort((a, b) => moment(a.dateCreated).unix() - moment(b.dateCreated).unix())
    .reduce((acc, transaction) => {
      const index = acc.findIndex(row => row.isCredit === !transaction.isCredit)
      index === -1 ? acc.push(transaction) : acc.splice(index, 1)
      return acc
    }, [])
}
```

StuAA78 (changes requested):
Great stuff, just spotted a few opportunities for some tweaks.

src/modules/billing/services/supplementary-billing-service/supplementary-processor.js
Beckyrose200: With those changes made it is showing as a code smell on Sonar Cloud. Ive reverted this file back to how it was , unless there is another way to do it :)

Cruikshanks (approved):

PR #2082, opened 2023-03-29
Remove 'flatMap' Lodash function

Beckyrose200:

Lodash is used a lot in our codebase. Although we are not totally removing this module, some of the functions in the code using lodash are unnecessary. Here we are removing the unnecessary ones are replacing it with vanilla js. This PR focuses on removing 'flatMap'.

### DIFF
--- a/src/modules/billing/controllers/batches.js
+++ b/src/modules/billing/controllers/batches.js
@@ -2,7 +2,6 @@
 
 const Boom = require('@hapi/boom')
 
-const { flatMap } = require('lodash')
 const { jobStatus } = require('../lib/event')
 const { createBatchEvent } = require('../lib/batch-event')
 const controller = require('../../../lib/controller')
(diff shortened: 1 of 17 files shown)


src/modules/billing/controllers/batches.js
```diff
@@ -174,13 +173,21 @@ const getBatchDownloadData = async request => {
...
+    invoices.map(invoice => {
+      // Get an array of billing invoice licences and filter out transactions
+      const filtered = invoice.billingInvoiceLicences.map(invoiceLicence =>
+        invoiceLicence.billingTransactions
+          .filter(transaction => !!transaction.chargeElement)
+          .map(transaction => transaction.chargeElement.chargeVersionId)
+      )
+      // Return a flattened array of charge version IDs
+      return filtered.flatMap(n => n)
+    }).flatMap(n => n)
```
StuAA78, line 188: If we replace this with `.flat(Infinity)` then we don't need the `.flatMap()` on line 187 as it will completely flatten the resulting array. But it might be worth replacing some or all of the other `.map()` in this with `.flatMap()`, which would hopefully give us a flat array without needing an extra 'and now flatten what we have' step.

StuAA78 (changes requested):
Today I learned that `.flatMap()` is a thing and we can use it in place of `.map(...).flat(Infinity)`! Just a few small tweaks we could maybe make.

StuAA78 (approved):

PR #2089, opened 2023-03-31
Remove 'reverse', 'compact', 'isFinite', 'pickBy' and 'orderBy' Lodash function

Beckyrose200:

Lodash is used a lot in our codebase. Although we are not totally removing this module, some of the functions in the code using lodash are unnecessary. Here we are removing the unnecessary ones are replacing it with vanilla js. This PR focuses on removing 'reverse', 'compact', 'isFinite', 'pickBy' and 'orderBy'.


src/modules/returns/lib/csv-adapter/date-parser.js
```diff
@@ -51,7 +52,8 @@ const getDateFrequency = date => {
     const m = parsers[timePeriod](date)
     return m.isValid() ? { timePeriod, moment: m } : null
   })
-  return compact(result)[0]
+
+  return result.filter(x => !!x)[0]
```
StuAA78, line 56: If I'm reading this correctly we're populating `result` with either a parsed date or `null`, then returning the first truthy value (ie. the first successful attempt at parsing). So instead of filtering out the falsey values then returning the first item in the array, we could use this to just return the first truthy value:
```suggestion
  return result.find(x => x)
```

src/modules/billing/services/charge-processor-service/lib/agreements.js
```diff
@@ -28,7 +28,7 @@ const mapAgreement = licenceAgreement => ({
 })
 
 const mapHistoryRow = row => {
-  const agreements = compact([row.section127Agreement, row.section130Agreement])
+  const agreements = [row.section127Agreement, row.section130Agreement].filter(x => !!x)
     .map(row => row.section127Agreement || row.section130Agreement)
```
StuAA78, line 32: Since we're already put `.map()` onto a separate line we probably want `.filter()` on its own line too (also swapping the `!!x` for plain `x` as mentioned earlier):
```suggestion
  const agreements = [row.section127Agreement, row.section130Agreement]
    .filter(x => x)
    .map(row => row.section127Agreement || row.section130Agreement)
```

src/lib/models/licence.js
```diff
@@ -112,9 +112,9 @@ class Licence extends Model {
    * @return {String|null}
    */
   get endDate () {
-    const dates = compact([this.expiredDate, this.lapsedDate, this.revokedDate])
-    const sorted = orderBy(dates, date => moment(date).unix())
-    return sorted[0] || null
+    const dates = [this.expiredDate, this.lapsedDate, this.revokedDate].filter(x => !!x)
```
StuAA78, line 115: I think here (and throughout) we may be able to just `.filter(x => x)` -- `!!x` turns `x` into a boolean to represent whether its truthy or falsey, but `.filter()` doesn't care about having an actual boolean -- it will simply use whether `x` is truthy or falsey.
```suggestion
    const dates = [this.expiredDate, this.lapsedDate, this.revokedDate].filter(x => x)
```
The only thing I'd say about this approach is that it will also filter out `0` from arrays; in most cases this is probably fine as `0` on its own would not be an expected entry but it's something to bear in mind!

StuAA78 (changes requested):

src/lib/services/invoice-service.js
```diff
@@ -124,7 +123,14 @@ const decorateInvoiceWithCRMData = (invoice, context) => {
     'company'
   )
 
-  return invoice.fromHash(pickBy(properties))
+  return invoice.fromHash(removeFalseyValues(properties))
+}
+
+const removeFalseyValues = (object) => {
+  // Convert the object into an array of key-value pairs, then filter out any entires with falsey values
+  const entries = Object.entries(object).filter(([_, value]) => !!value)
```
StuAA78, line 131: Does this need to be `!!value` or would `value` work as well?

src/modules/batch-notifications/lib/notify-helpers.js
```diff
@@ -7,7 +7,7 @@ const { compact } = require('lodash')
  * @return {Array}
  */
 const reduceAddressLines = lines => {
-  const compacted = compact(lines)
+  const compacted = lines.filter(x => !!x)
```
StuAA78, line 10: Does this need to be `!!x` or would `x` work as well?

src/modules/returns/lib/csv-adapter/date-parser.js
```diff
@@ -51,7 +52,8 @@ const getDateFrequency = date => {
     const m = parsers[timePeriod](date)
     return m.isValid() ? { timePeriod, moment: m } : null
   })
-  return compact(result)[0]
+
+  return result.filter(x => x)[0]
```
StuAA78, line 56: Does `result.first(x => x)` work here?

src/modules/returns/lib/csv-adapter/validator.js
```diff
@@ -132,7 +132,7 @@ const validateHeadings = records => {
     return acc
   }, [])
 
-  return compact(errors)
+  return errors.filter(x => !!x)
```
StuAA78, line 135: Does this need to be `!!x` or would `x` work as well?

src/modules/returns/lib/csv-adapter/validator.js
```diff
@@ -237,7 +237,7 @@ const validateLicences = licences => {
     return acc
   }, [])
 
-  return flatten(compact(errors))
+  return flatten(errors.filter(x => !!x))
```
StuAA78, line 240: Does this need to be `!!x` or would `x` work as well?

StuAA78 (changes requested):
Spotted a few places where we're using `!!x` etc. when `x` may also work; and I wasn't sure if the `return result.filter(x => x)[0]` was intended to remain like that or if you'd tried `return result.first(x => x)`

src/modules/returns/lib/csv-adapter/date-parser.js
Beckyrose200: Changing this to .first doesn't work as its complaining its not a function `  result.first is not a function` 😢

src/lib/services/invoice-service.js
Beckyrose200: These changes all worked 😄

Cruikshanks (dismissed):

src/modules/returns/lib/csv-adapter/date-parser.js
StuAA78: Sorry my mistake! I meant `result.find(x => x)` 🙄

StuAA78 (changes requested):
Just the one bit which was my mistake -- I left the comment referring to `result.first(x => x)` when I meant to put `result.find(x => x)`!

Cruikshanks (approved):

Jozzey (approved):
👍🏼

StuAA78 (approved):

PR #2093, opened 2023-04-04
Remove 'groupBy' and 'mapValues' Lodash function

Beckyrose200:

Lodash is used a lot in our codebase. Although we are not totally removing this module, some of the functions in the code using lodash are unnecessary. Here we are removing the unnecessary ones are replacing it with vanilla js. This PR focuses on removing 'groupBy' and 'mapValues'.


Cruikshanks (approved):

src/modules/notify/connectors/scheduled-notification.js
```diff
@@ -2,14 +2,19 @@
...
 const { findOne } = require('../../../lib/repository-helpers')
 
 const stringifyArray = (value) => Array.isArray(value) ? JSON.stringify(value) : value
 
 const mapObjectToNotification = (data) => {
-  return mapValues(snakeCaseKeys(data), stringifyArray)
+  const snakeCaseData = snakeCaseKeys(data)
+  const dataStringify = {}
+  for (const data in snakeCaseData) {
+    dataStringify[data] = stringifyArray(snakeCaseData[data])
```
StuAA78, line 14: If `stringifyArray` is only used in one place then do we need it to be a separate function or can we inline it, eg:
```suggestion
    const value = snakeCaseData[data]
    dataStringify[data] = Array.isArray(value) ? JSON.stringify(value) : value
```

StuAA78 (changes requested):

PR #2116, opened 2023-04-21
Supplementary Billing - Too many previous years

Jozzey:
https://eaflood.atlassian.net/browse/WATER-3944

A bug has been found in calculating the number of years to go back for ALCS supplementary billing.

The issue was caused because the `toFinancialYearEnding` variable was being modified for alcs if it was greater than 2022. Unfortunately, this value was then being used to calculate the number of years to process for alcs which was incorrect as the variable's value had been changed to 2022 prior to this step.

This fix moves the step of modifying the variable to below where it is used to calculate the number of financial years to process.


Cruikshanks (approved):
Blinding find!

PR #2178, opened 2023-06-12
Removing SRoC flag sets Pre-SRoC supplementary flag

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4040

Removing the flag from a licence for a SRoC bill run puts the Pre-SRoC supplementary flag on the licence.


Cruikshanks (approved):
🤦 Obviously we need to keep `await deleteInvoicesWithRelatedData(batch, invoice)` in 😝 

Sorry for the bad tip!

PR #2284, opened 2023-09-21
Migration to delete broken charge version

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4147

This is a migration to delete a broken charge version record that doesn't have a start date.


migrations/sqls/20230921102640-delete-broken-charge-versions-up.sql
```diff
@@ -0,0 +1,9 @@
+/* deletes charge versions in the workflow that have incorrect data */
+CREATE TEMP TABLE tempChargeVersionId
+AS (
+    SELECT charge_version_workflow_id AS id
+    FROM water.charge_version_workflows
+    WHERE charge_version_workflow_id = '4fe385f1-a50c-4608-91ee-96a3b15ad3b6'
+);
+DELETE FROM water.charge_version_workflows WHERE charge_version_workflow_id = '4fe385f1-a50c-4608-91ee-96a3b15ad3b6';
+DROP TABLE tempChargeVersionId;
```
Cruikshanks, line 9: The query this was copied from relied on selecting IDs based on a WHERE filter. We know the ID in this case so we can drop a lot of what was there.

```suggestion
DELETE FROM water.charge_version_workflows WHERE charge_version_workflow_id = '4fe385f1-a50c-4608-91ee-96a3b15ad3b6';

```

migrations/sqls/20230921102640-delete-broken-charge-versions-down.sql
```diff
@@ -0,0 +1 @@
+/* Replace with your SQL commands */
```
Cruikshanks, line 1: I blame the migration tool. But because it doesn't finish with a newline GitHub puts a 'shouty' red circle at the end to warn you.

```suggestion
/* Replace with your SQL commands */

```

Cruikshanks (changes requested):
Just a tweak to remove some 'deadwood' from the query!

migrations/sqls/20230921102640-delete-broken-charge-versions-up.sql
Beckyrose200: I thought it would be that simple but I assumed they created a temp table to keep the IDs in case it failed 😬

migrations/sqls/20230921102640-delete-broken-charge-versions-down.sql
Cruikshanks: Looking at the changes we still don't have a new line at the end of this file. GitHub is still showing

Screenshot 2023-09-21 at 12 55 01

If you commit my suggestion it will sort this. I also find that having `Insert Final Newline` ticked in my VSCode settings means I never commit a file without one.

Screenshot 2023-09-21 at 12 56 26

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #2288, opened 2023-09-26
Prevent multiple 'send a ready bill run'

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4100

We're making changes to our job queue system to prevent the same bill run from being processed more than once. Previously, we fixed an issue by removing failed jobs from the queue (https://github.com/DEFRA/water-abstraction-service/pull/2205), which allowed us to add a new bill run with the same ID. However, this created a problem where multiple identical bill runs could be added if someone clicked the "send bill run" button twice quickly.

If someone clicks the "send bill run" button twice quickly, the first click adds the bill run, but the second click removes it and adds it again. If the bill run is removed from the queue by a worker, the second bill run with the same ID remains in the queue because it's the only one with that ID.

To solve this problem, we've stopped removing jobs whenever a bill run is sent. Now, our job queue system can recognize when there are multiple bill runs with the same ID and prevent them from being added. To ensure that bill runs don't get stuck, we've only added the job.remove() method to the failedHandler function.


Cruikshanks (changes requested):
Just some typos in the comments. It also makes it sound like we added the remove method when in fact this is something BullMQ attaches to the job object.

Jozzey (commented):
Nothing to add, looks good 👍🏼

Cruikshanks (approved):

PR #2295, opened 2023-10-04
Improve Hand-Off Flow Notification Handling

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4110

This pull request is working on preventing the system from sending multiple hand-off flow-level warning letters. Initially, it was assumed that each notification letter/email operated independently. Upon investigation, it was discovered that a single generic notification system was used, utilizing only one notification template with personalised data. 

This constraint meant that a straightforward solution like restricting the BullMq queue to a single item wouldn't work. The data was also added to the database before job completion, initially marked as "draft," and subsequently updated to "sent" upon successfully sending the notification. This was done to maintain a record in case of any issues with the notify API, enabling the possibility of resending. In the table, there is a column called message_ref which is meant to indicate the type of notification being sent. 

Due to the fact that all four notification types (1: Restriction, 2: Renewal letters, 3: Hands-off flow, 4: Resume) follow the same process, they are collectively labeled as 'notification' in the database message_ref column. To address this, a new column has been introduced in the scheduled_notifications table to record the notification type. Either 1,2,3 or 4. 

To prevent the processing of multiple requests, we introduced a new "job_id" column within the "scheduled_notifications" table. Once a user inputs notification data and moves to the preview page, a unique ID is generated within the UI and gets passed to our service. When a notification request is received, this ID is utilized as the BullMQ job ID. If a duplicate request emerges, it will possess the same ID, prompting BullMQ to disregard it.

This solved the problem of multiple notifications being sent. However, we still encountered an issue where the record was added to the database for each request because this insertion occurred before the task was enqueued. To address this, we introduced a constraint on the "job_id" column, allowing only unique values to be inserted. As a result, any subsequent attempts to insert the same record are rejected by the database, ensuring uniqueness.

### DIFF
--- a/src/modules/notifications/lib/index.js
+++ b/src/modules/notifications/lib/index.js
@@ -76,6 +76,7 @@ async function sendNotification (queueManager, taskConfig, issuer, contactData, 
   // Schedule messages for sending
   for (const row of contactData) {
     const n = await notificationFactory(row, taskConfig, e)
+    n.notificationType = taskConfig.task_config_id
 
     try {
       await enqueue(queueManager, n)
--- a/src/modules/notifications/lib/index.js
+++ b/src/modules/notifications/lib/index.js
@@ -74,16 +74,14 @@ async function sendNotification (queueManager, taskConfig, issuer, contactData, 
   logger.info(`Sending notification reference ${e.referenceCode} ID ${e.eventId}`)
 
   // Schedule messages for sending
-  let rowCount = 0
-  for (const row of contactData) {
+  for (const [i, row] of contactData.entries()) {
     const n = await notificationFactory(row, taskConfig, e)
     n.notificationType = taskConfig.task_config_id
-    const rowJobId = `${uniqueJobId}${rowCount}`
-    rowCount++
+    const rowJobId = `${uniqueJobId}${i}`
     n.jobId = rowJobId
 
     try {
-      await enqueue(queueManager, n, rowJobId)
+      await enqueue(queueManager, n)
     } catch (error) {
       logger.error('Error sending notification', error.stack)
     }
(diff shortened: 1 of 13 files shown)


src/modules/notifications/lib/index.js
```diff
@@ -66,19 +66,24 @@ async function prepareNotification (filter, taskConfig, params, context = {}) {
...
-async function sendNotification (queueManager, taskConfig, issuer, contactData, ref) {
+async function sendNotification (queueManager, taskConfig, issuer, contactData, ref, uniqueJobId) {
   // Create event
   const e = eventFactory(issuer, taskConfig, contactData, ref)
   await evt.save(e)
 
   logger.info(`Sending notification reference ${e.referenceCode} ID ${e.eventId}`)
 
   // Schedule messages for sending
+  let rowCount = 0
```
Cruikshanks, line 77: If you are interested, there is a way we can get an index and not have to rely on a `rowCount` variable 🤷. Easier to demo rather than submit a suggestion though 😁

Cruikshanks (commented):
I have a suggestion to handle getting the row count in a different way we can take a look at.

But the key thing is SonarCloud is right on this one. So, even if we disregard the rowcount idea we should pair on getting the SonarCloud issue.

Just gonna mark this for 'comment' now as I'm not 💯 confident about resolving that SonarClopud issue in this legacy code 😩 😂

src/modules/notifications/lib/index.js
```diff
@@ -66,16 +66,19 @@ async function prepareNotification (filter, taskConfig, params, context = {}) {
...
 
   logger.info(`Sending notification reference ${e.referenceCode} ID ${e.eventId}`)
 
   // Schedule messages for sending
-  for (const row of contactData) {
+  for (const [i, row] of contactData.entries()) {
     const n = await notificationFactory(row, taskConfig, e)
+    n.notificationType = taskConfig.task_config_id
+    const rowJobId = `${uniqueJobId}${i}`
+    n.jobId = rowJobId
```
Cruikshanks, line 81: Any objections to doing this?

```suggestion
    n.jobId = `${uniqueJobId}${i}`
```

Cruikshanks (approved):

PR #2320, opened 2023-10-25
Consolidate Returns Letters and remove unused Code

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4185

This PR aims to eliminate redundant notify templates. Currently, within the code for return invitations and reminders, more than 32 templates are in use. Many of these templates share the same content. However, because the code currently selects from any of these 32 templates, it necessitates keeping all of them up to date. This change seeks to rectify this issue by reducing the number of templates in use. Specifically, we have identified 4 template IDs for notify that should be used for reminders and another 4 for invitations. Any references to the other templates can be safely removed. Additionally, this PR also eliminates the logic responsible for determining which template to use for invitations, as it was the same template selected for reminders.

Note: During this pull request, we didn't know how the "scheduled_notifications_categories" table was used in the code. I created a migration for this table to get rid of the extra templates, but as a result, the search function for filtering notifications stopped working. Specifically, when filtering for return invitations and reminders, it relies on matching "scheduled_notification_refs" to the "message_ref" of the notification. Removing the old templates means they won't appear in the filtered list of notifications for the user.


Jozzey (commented):
Looks alright to me

Cruikshanks (approved):

PR #2328, opened 2023-11-02
Delete bugged licence

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4182

This PR contains the migrations to remove a licence that is containing a carriage return at the end of the licence ref. This is causing the UI to error and our acceptance tests to fail.

### DIFF
--- a/migrations/sqls/20231102122939-delete-broken-licence-up.sql
+++ b/migrations/sqls/20231102122939-delete-broken-licence-up.sql
@@ -29,7 +29,7 @@ DELETE FROM water.licence_version_purposes WHERE licence_version_id IN (
 
 DELETE FROM water.licence_versions WHERE licence_id IN (
   SELECT licence_id FROM water.licences WHERE licence_ref LIKE 'NW/072/0417/002/R01' || CHR(13)
-)
+);
 
 DELETE FROM water.return_requirement_purposes WHERE return_requirement_id IN (
   SELECT return_requirement_id  FROM water.return_requirements WHERE return_version_id IN (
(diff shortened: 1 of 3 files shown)


migrations/sqls/20231102122939-delete-broken-licence-up.sql
```diff
@@ -0,0 +1,112 @@
...
+    BEGIN
+        IF EXISTS
+            ( SELECT 1
+              FROM   information_schema.tables
+              WHERE  table_schema = 'crm_v2'
+              AND    table_name = 'document_roles'
+            )
+        THEN
+          DELETE FROM "crm_v2"."document_roles" WHERE document_id IN (
+            SELECT document_id FROM "crm_v2"."documents" WHERE document_ref LIKE 'NW/072/0417/002/R01' || CHR(13)
```
Jozzey, line 12: This is my bad as I was on the call with you when you worked on this but I think where you have used `WHERE licence_ref LIKE 'NW/072/0417/002/R01' || CHR(13)` you can use an `=` instead of `LIKE` so it's `WHERE licence_ref = 'NW/072/0417/002/R01' || CHR(13)`. That should work for all of them

Jozzey (changes requested):
Just 1 comment. Sorry I should have spotted that when we were trying to figure out how to search for that carriage return.

migrations/sqls/20231102122939-delete-broken-licence-up.sql
Beckyrose200: All changed :)

Beckyrose200: All changed :)

Cruikshanks (approved):

PR #2346, opened 2023-11-22
Add defaults to unused `returns` & `lines` table columns

Jozzey:
https://github.com/DEFRA/water-abstraction-team/issues/106

As part of making the migrations work in `water-abstraction-system` we have been creating DB Views of the legacy Returns data that better represent the data as we use it in our new service. Part of this exercise is pruning out the unused columns that exist in the legacy tables that are not utilised.

However, we have found that for some of the columns in the legacy data, whilst pointless, still need to be populated as they are set to Not Nullable in the legacy database tables. Therefore, we will set a default value for these columns in the legacy DB so that they are always populated even when they do not exist in our new Views.

The table columns that will be affected by this PR are:
- `returns.regime` - this is always "water" so it will default to "water"
- `returns.licence_type` - this is always "abstraction" so it will default to "abstraction"
- `lines.substance` - surprise, surprise, this is always "water" so it will default to "water"
- `lines.unit` - This is always "m³" so it will default to "m³"

### DIFF
--- a/migrations/20231122170046-add-defaults-to-returns-tables.js
+++ b/migrations/20231122170046-add-defaults-to-returns-tables.js
@@ -1,0 +1,45 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+let Promise
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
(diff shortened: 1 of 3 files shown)


Jozzey:
I created the migrations in the wrong repo. Should have created them in the Returns repo

PR #2381, opened 2024-01-04
Fix missing invoice account address records

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4223

We are replacing the legacy bill run view with one we have developed in [water-abstraction-system](https://github.com/DEFRA/water-abstraction-system). During testing, it was found that there are examples of bill runs where a bill is missing from the view. When we checked the legacy version the bill was present so this was an issue just in our version of the page.

When we looked into the reason why, we found it was because the bill (`water.billing_invoice`) was linked to a billing account (`crm_v2.invoice_account`) that has no account address record (`crm_v2.invoice_account_address`). Since starting with the service we have been working under the assumption this should _never_ be the case; each billing account should have at least **one** account address record.

The legacy code is catering for this but when we checked we found only 38 records were affected. So, rather than add complexity to our code to handle 'duff' data it would be better to fix the root cause.

So, this fix migration creates the missing `crm_v2.invoice_account_address` records for the billing accounts affected.

### DIFF
--- a/migrations/20240104130338-fix-missing-invoice-account-addresses.js
+++ b/migrations/20240104130338-fix-missing-invoice-account-addresses.js
@@ -1,0 +1,45 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+let Promise
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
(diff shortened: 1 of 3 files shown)


Cruikshanks:
Wrong service you numbty Cruikshanks! See https://github.com/DEFRA/water-abstraction-tactical-crm/pull/1005

PR #2386, opened 2024-01-11
Removing console.log() from legacy migrations

rvsiyad:
https://github.com/DEFRA/water-abstraction-team/issues/91

This PR is focuing on removing console.log() from legacy migrations as per Teams issue.


Cruikshanks (approved):

Cruikshanks:
Just a note about the PR description @rvsiyad (because it then becomes the squashed commit message). We like to link to the source item (teams issue or Jira ticket) as it might give a bit more context for anyone taking a look when reviewing.

But the source is 'king' and the commit history is what you'll immediately see in your editor when looking at a line of code. So, even if it means duplicating content from the ticket, treat the PR description/squashed commit message as the place to detail _why_ a change is being made.

PR #2455, opened 2024-03-08
add display_title column to licence_version_purpose_condition_types 

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4332

As part of the work to migrate the licence summary page to the new water-abstraction-system ui we've found that the backend service is using a csv file to get the display titles for abstraction conditions. 

This PR is to add display_title column to licence_version_purpose_condition_types and populate it with the values in that csv file.


Cruikshanks (changes requested):
Just missing some newlines at the end of the SQL files. It avoids GitHub displaying this.

Screenshot 2024-03-14 at 10 37 40

If it helps running **🌈 Settings** from [wal-dev-environment](https://gitlab-dev.aws-int.defra.cloud/water-abstraction/wal-dev-environment/-/blob/main/docs/vscode/README.md#settings) _should_ copy settings to this repo that will do this automatically for you on save.

Or checkout details for setting [directly](https://stackoverflow.com/a/44704969/6117745)

Also, before I hit approve could I ask you add a bit more detail to the PR description (and therefore the commit message on squash) as to _why_ we are doing this? Just a couple of sentences related to your work on viewing a licence and that this change means we can access the data we need easier than the way the legacy service was doing it.

Cruikshanks (approved):

PR #2523, opened 2024-05-13
Handle rtn. req. not existing when mapping rtns

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4465

We had a call that when attempting to view a licence the users were seeing an error. Initially, we thought it was the usual issue of a second record with leading or trailing whitespace. But the problem was actually missing data.

When the view licence page is loading the legacy code is fetching all return logs for a licence (though the page only displays the first ten results we suspected this was the case!)

What we were not aware of is it then tries to map each return log returned to a record in `water.return_requirements`. Nothing from that record is displayed as part of the view licence page and as far as we know, no functionality uses it.

But for reasons we don't know, the records are imported and the returns mapper in this service attempts to find the matching return requirement!

Anyway, when we first joined the team we had it in mind to start improving and resolving the many issues we found in the legacy code. One of these was the use of Lodash for pretty much anything!

Its [find()](https://lodash.com/docs/4.17.15#find) was being used in the mapper to identify the matching return requirement. The code as written appears to assume there will always be a result.

So, as part of our efforts to ditch Lodash, we replaced `find()` with the [standard JavaScript find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find).

All seemed well until the users hit on the one licence where there isn't a return requirement for every return log! The Lodash `find()` will silently eat the error. The standard library will throw if no result is found.

Rather than go back to Lodash, this change wraps the call in a try/catch. We now know it doesn't matter if the function returns undefined. Testing has confirmed the page will load. So we can catch the exception, log the fact this has happened and then let the function continue as it did before.

We are only a couple of weeks away from replacing the legacy view licence page. So, we see this as a stop-gap solution until then anyway.

### DIFF
--- a/src/lib/services/returns/returns-mapping-service.js
+++ b/src/lib/services/returns/returns-mapping-service.js
@@ -42,7 +42,6 @@ const mapReturnDataToModel = (returnData, returnRequirements) => {
 
   try {
     returnRequirement = returnRequirements.find((o) => o.externalId === externalId)
-    // returnRequirement = returnRequirements.find((o) => o.externalId === externalId)
   } catch (error) {
     logger.error(`Error finding return requirement ${externalId} in mapping service`, error.stack)
   }
--- a/src/lib/services/returns/returns-mapping-service.js
+++ b/src/lib/services/returns/returns-mapping-service.js
@@ -44,7 +44,7 @@ const mapReturnDataToModel = (returnData, returnRequirements) => {
     returnRequirement = returnRequirements.find((o) => o.externalId === externalId)
     // returnRequirement = returnRequirements.find((o) => o.externalId === externalId)
   } catch (error) {
-    logger.error('Error finding return requirement in mapping service', error.stack)
+    logger.error(`Error finding return requirement ${externalId} in mapping service`, error.stack)
   }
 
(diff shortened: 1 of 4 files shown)


src/lib/services/returns/returns-mapping-service.js
```diff
@@ -35,7 +37,16 @@ const getReturnRequirements = returnsData => {
  */
 const mapReturnDataToModel = (returnData, returnRequirements) => {
   const externalId = getReturnRequirementExternalId(returnData)
-  const returnRequirement = returnRequirements.find((o) => o.externalId === externalId)
+
+  let returnRequirement
+
+  try {
+    returnRequirement = returnRequirements.find((o) => o.externalId === externalId)
+    // returnRequirement = returnRequirements.find((o) => o.externalId === externalId)
```
Jozzey, line 45: Did you mean to leave this commented-out line here?

Jozzey (changes requested):

src/lib/services/returns/returns-mapping-service.js
Cruikshanks: Doh!

Jozzey (approved):

PR #2529, opened 2024-05-18
Add reason column to water.return_versions table

robertparkinson:
https://eaflood.atlassian.net/browse/WATER-4470

The new return requirements setup journey captures a reason for the requirement, and we don't currently have a place to store it. This change adds a column to the `water.return _versions` table to store it.

### DIFF
--- a/migrations/20240518095436-water-return-versions-add-reason-column.js
+++ b/migrations/20240518095436-water-return-versions-add-reason-column.js
@@ -1,0 +1,45 @@
+'use strict'
+
+const fs = require('fs')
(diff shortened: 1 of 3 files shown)


Cruikshanks (approved):

Cruikshanks:
I hope you don't mind, @robertparkinson . But I also have a change in this repo which is _also_ a migration. What are the chances!?

Using my Tech Lead powers I need to get mine in, but I also didn't want to cause any problems with the ordering of the scripts.

So, I've merged your PR in so that any problems are my fault!

PR #2600, opened 2024-08-12
Add System config for Supplementary billing flags

Beckyrose200:
https://eaflood.atlassian.net/browse/WATER-4622

As part of the work for two-part tariff supplementary billing, we need to add the system config for the service repo to hit a new endpoint we have created. The endpoint is to flag a licence for supplementary billing. As there are 5 different ways a licence can get flagged with 3 of them triggered from the service repo, this PR will allow us to hit the system endpoint and add the flag if necessary.

### DIFF
--- a/src/lib/connectors/system/supplementary-billing.js
+++ b/src/lib/connectors/system/supplementary-billing.js
@@ -4,7 +4,7 @@ const config = require('../../../../config')
 const { serviceRequest } = require('@envage/water-abstraction-helpers')
 
 /**
- * Posts to
+ * Posts to system repo to flag new charge versions for SROC supplementary billing
  * @return {Promise}
  */
 const flagSupplementaryBilling = (chargeVersionId) => {
--- a/src/lib/connectors/system/supplementary-billing.js
+++ b/src/lib/connectors/system/supplementary-billing.js
@@ -1,0 +1,22 @@
+'use strict'
+
+const config = require('../../../../config')
+const { serviceRequest } = require('@envage/water-abstraction-helpers')
+
+/**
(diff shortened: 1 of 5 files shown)


src/lib/connectors/system/supplementary-billing.js
```diff
@@ -0,0 +1,22 @@
...
+
+const config = require('../../../../config')
+const { serviceRequest } = require('@envage/water-abstraction-helpers')
+
+/**
+ * Posts to
+ * @return {Promise}
+ */
+const flagSupplementaryBilling = (chargeVersionId) => {
+  console.log('Charge Version Id :', chargeVersionId)
```
Cruikshanks, line 11: Need to remove the `console.log()`.

src/lib/services/charge-versions.js
```diff
@@ -104,6 +105,11 @@ const persist = async chargeVersion => {
   )
   persistedChargeVersion.chargeElements = await Promise.all(tasks)
 
+  console.log('Charge Version', persistedChargeVersion)
+  console.log('AHHHH Im in service!!!!')
```
Cruikshanks, line 109: Need to remove the `console.log()`s.

Cruikshanks (changes requested):
The repo also has a `.env.example` file that should be updated with the new config.

src/lib/services/charge-versions.js
Beckyrose200: I didn't realise this was up for review sorry, not finished with it!

Cruikshanks (approved):

PR #2603, opened 2024-08-13
Treat supplementary billing as a background job

Beckyrose200:
https://github.com/DEFRA/water-abstraction-service/pull/2600
https://eaflood.atlassian.net/browse/WATER-4622
https://eaflood.atlassian.net/browse/WATER-4588

During some feedback on the supplementary billing flag services created, we decided the flagging service should be treated as a background job that gets kicked off but doesn't need to await any results. Hence this PR is to remove the await before the call to the system repo.

### DIFF
--- a/src/lib/services/charge-versions.js
+++ b/src/lib/services/charge-versions.js
@@ -105,7 +105,7 @@ const persist = async chargeVersion => {
   )
   persistedChargeVersion.chargeElements = await Promise.all(tasks)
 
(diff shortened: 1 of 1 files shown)


Cruikshanks (changes requested):
Eek! Something has got lost in translation. You _need_ the await here. **water-abstraction-service** has to confirm its request was received. With the await removed, if the request failed it would crash the app.

So this PR needs to be cancelled 😬

PR #2605, opened 2024-08-14
Add migration for licence versions and licence versions purposes

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4579

Alters the water.licence_versions and water.licence_version_purposes to set the date_created and date_updated to default now.

This has been done as part migrating the legacy import licence service to the water-abstraction-system. Rather than adding this logic in the system code it makes sense to add the migration and have the database handle this logic for us.

### DIFF
--- a/migrations/sqls/20240814102304-alter-licence-version-created-and-updated-at-down.sql
+++ b/migrations/sqls/20240814102304-alter-licence-version-created-and-updated-at-down.sql
@@ -1,6 +1,6 @@
 BEGIN;
 
-ALTER TABLE IF EXISTS water.licence_versions ALTER COLUMN date_created SET NOT NULL;
-ALTER TABLE IF EXISTS water.licence_versions ALTER COLUMN date_updated SET NOT NULL;
+ALTER TABLE IF EXISTS water.licence_versions ALTER COLUMN date_created DROP DEFAULT;
+ALTER TABLE IF EXISTS water.licence_versions ALTER COLUMN date_updated DROP DEFAULT;
 
 COMMIT;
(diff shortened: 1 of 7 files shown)


migrations/sqls/20240814102304-alter-licence-version-created-and-updated-at-down.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE IF EXISTS water.licence_versions ALTER COLUMN date_created SET NOT NULL;
+ALTER TABLE IF EXISTS water.licence_versions ALTER COLUMN date_updated SET NOT NULL;
```
Cruikshanks, line 4: This is not my understanding on how to remove a default from a column. I think instead it's supposed to be

```suggestion
ALTER TABLE IF EXISTS water.licence_versions ALTER COLUMN date_created DROP DEFAULT;
ALTER TABLE IF EXISTS water.licence_versions ALTER COLUMN date_updated DROP DEFAULT;
```

Having run the migration, if you run the migration down command locally you should see the column properties revert to what they were before the migration was applied.

migrations/sqls/20240814103242-alter-licence-version-purposes-created-and-updated-at-down.sql
```diff
@@ -0,0 +1,7 @@
+/* Replace with your SQL commands */
+BEGIN;
+
+ALTER TABLE IF EXISTS water.licence_version_purposes ALTER COLUMN date_created SET NOT NULL;
+ALTER TABLE IF EXISTS water.licence_version_purposes ALTER COLUMN date_updated SET NOT NULL;
```
Cruikshanks, line 5:
```suggestion
ALTER TABLE IF EXISTS water.licence_version_purposes ALTER COLUMN date_created DROP DEFAULT;
ALTER TABLE IF EXISTS water.licence_version_purposes ALTER COLUMN date_updated DROP DEFAULT;
```

Cruikshanks (changes requested):
Not sure the command is right for the down queries.

Cruikshanks (approved):

PR #2606, opened 2024-08-14
Add new `batch_type` for TPT supplementary to DB

Jozzey:
https://eaflood.atlassian.net/browse/WATER-4200

In order to enable the triggering of a two-part tariff supplementary bill run we first have to create a new data type for it in our DB.

In this PR the data type is going to be updated to add a new batch type of `tpt_supplementary`. Any issues that adding this new data type may cause will also be actioned in this PR.

### DIFF
--- a/migrations/20240814134405-alter-water-charge-batch-type.js
+++ b/migrations/20240814134405-alter-water-charge-batch-type.js
@@ -1,0 +1,45 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+let Promise
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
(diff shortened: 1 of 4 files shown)


Jozzey:
It's only possible to create or alter ENUM values, you can't delete a value once created. So it's not possible to have a down script for the migration. Because of this, the migration script checks to see if the new value already exists before trying to add it. If `IF NOT EXISTS` was not added to the migration script, and `npm run migrate:down` was run, followed by `npm run migrate` it would break.

Jozzey:
I've searched through the legacy code and can't find anywhere that the addition of the new batch type to the DB should cause any problems.

Cruikshanks (approved):

PR #2609, opened 2024-08-15
Fix external change user email journey

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4632

We received a report that the external app's change email journey was broken. We quickly confirmed this.

The [first part of the fix](https://github.com/DEFRA/water-abstraction-tactical-idm/pull/886) was some mis-ordered logic in the **water-abstraction-tactical-idm** project.

After spotting this, we thought we were done. However, the feature was breaking at the point you entered the security code.

We tracked it down to an issue in this project. It seems we create a `water.event` record upon confirmation of the change being verified. Or we would if it didn't blow up.

We first spotted that it was using a deprecated method for creating the event (the method and the deprecation were all created by the previous team!)

It also saved the record using [Bookshelf](https://bookshelfjs.org). By this point, we were a little fed up, so forgive us for not digging into the history.

But we found the problem was the helper was generating an event object with an `event_id`. That event object was then getting passed to [Bookshelf's save() method](https://bookshelfjs.org/api.html#Model-instance-save). In **Bookshelf** `save()` is used for both updates and inserts. It seems **Bookshelf** determines which by checking if the thing it's saving has an ID. If it does, it uses an update. No ID, then it uses insert.

Setting the ID in the object was making **Bookshelf** think it was doing an update. However, when the DB returned 0 for the update count **Bookshelf**, it was blowing up. Much like [the fix](https://github.com/DEFRA/water-abstraction-tactical-idm/pull/886), this appears to have been a broken journey from the start.

In this change, we first switch to the non-deprecated process for creating event records. We then ensure that the helper doesn't generate an ID, which means **Bookshelf** makes an insert call instead.

All seems fine locally, but we're adding a [new external user change email test](https://github.com/DEFRA/water-abstraction-acceptance-tests/pull/131) just to be sure.

### DIFF
--- a/test/modules/change-email/controller.test.js
+++ b/test/modules/change-email/controller.test.js
@@ -14,7 +14,8 @@ const idm = require('../../../src/lib/connectors/idm')
 const crmEntities = require('../../../src/lib/connectors/crm/entities')
 const notifications = require('../../../src/lib/notifications/emails')
 const { logger } = require('../../../src/logger')
-const event = require('../../../src/lib/event')
+// const event = require('../../../src/lib/event')
+const Event = require('../../../src/lib/services/events.js')
 
 const userId = 123
 const email = 'mail@example.com'
@@ -34,7 +35,7 @@ experiment('change email controller', () => {
     sandbox.stub(notifications, 'sendEmailAddressInUseNotification')
     sandbox.stub(crmEntities, 'updateEntityEmail')
     sandbox.stub(logger, 'error')
-    sandbox.stub(event, 'save')
+    sandbox.stub(Event, 'create')
   })
 
   afterEach(async () => sandbox.restore())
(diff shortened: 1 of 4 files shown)


test/modules/change-email/controller.test.js
```diff
@@ -14,7 +14,8 @@ const idm = require('../../../src/lib/connectors/idm')
 const crmEntities = require('../../../src/lib/connectors/crm/entities')
 const notifications = require('../../../src/lib/notifications/emails')
 const { logger } = require('../../../src/logger')
-const event = require('../../../src/lib/event')
+// const event = require('../../../src/lib/event')
```
robertparkinson, line 17: did you mean to leave this commented out or included?

robertparkinson (changes requested):

test/modules/change-email/controller.test.js
Cruikshanks: Nope! Good spot!

robertparkinson (approved):

PR #2643, opened 2024-10-03
Data fix for multiple 'current' licence holders

Cruikshanks:
https://eaflood.atlassian.net/browse/WATER-4696

Specifically, `7/34/11/*G/0462` needs the fix. Users reported an issue with licence holder information.

- Two licence holders were listed on the view licence contact details page (there should only be one)
- When adding a new charge version, the journey was suggesting to create the new billing account using the old company details

We found the issue was that for this licence, we have two entries in `crm_v2.document_roles` with the role of `licenceHolder` and no end dates. In the case of the licence holder, the service derives the 'current' licence holder by selecting the one with no end date (there should only be one).

Because we have 2 with no end date, view contact details shows both. In the legacy charge version journey, it gets all licence roles for the licence, filters out any that are not 'licenceHolder', and then selects the first that meets this criteria.

- Start date is before or the same as the chosen date for the new charge version
- End date is null or after the chosen date for the new charge version

Again, the old licence holder is the first record to meet these criteria; hence, the legacy code chooses it as the 'relevant company' to display.

We believe the 'extra licence holder' results from a licence version in NALD but has since been deleted. We now know that [water-abstraction-import](https://github.com/DEFRA.water-abstraction-import) never deletes records on our side that no longer exist in NALD. So, this likely has become orphaned because its source record no longer exists in NALD.

### DIFF
--- a/migrations/20241003162404-fix-multiple-current-licence-holders.js
+++ b/migrations/20241003162404-fix-multiple-current-licence-holders.js
@@ -1,0 +1,45 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+let Promise
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function (options, _seedLink) {
+  Promise = options.Promise
+}
+
+exports.up = function (db) {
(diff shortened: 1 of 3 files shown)


Cruikshanks:
Doh! Wrong repo! This should have been created in [water-abstraction-ta](https://github.com/DEFRA/water-abstraction-tactical-crm/pull/1052) as owner of the `crm_v2` schema.

PR #2653, opened 2024-11-07
Add quarterly returns to return versions table

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4707

As part of the work to introduce quarterly returns in the system repo we need to add a column to mark a return version for quarterly returns.

The work to handle this logic is done in the system repo.

This change is just to introduce the quarterly returns in the previous water tables.


migrations/sqls/20241107093135-add-quarterly-returns-to-return-versions-up.sql
```diff
@@ -0,0 +1,2 @@
+alter table water.return_versions
+    add column quarterly_returns boolean default false;
```
Cruikshanks, line 2: As a boolean, unless there is an explicit need IMHO it is also good practise to make it `NOT NULL`.

```suggestion
    add column quarterly_returns boolean not null default false;
```

Cruikshanks (changes requested):

Cruikshanks (approved):

PR #2656, opened 2024-11-20
Spike upgrading to node v22

jonathangoulding:
https://eaflood.atlassian.net/browse/WATER-4729

We want to upgrade to node v22, but we have a blocker with the xml parser we are using.

The tool being used is libxmljs and has some open issues:

https://github.com/libxmljs/libxmljs/issues/664
https://github.com/libxmljs/libxmljs/issues/660

Mainly the issues are:
- is it being maintained
- is there going to be a release to make it work with node v22.

Rebuilding the package locally during our npm install has not worked, (through minimum effort) and if we did make it to build there is no guarantee it will actually work when built.

This spike investigates the effort required to get any xml parser to work with the least possible effort.

The change suggest using libxmljs2. Which works.

However, if you look at the documentation it states : "LibXML bindings for node.js This package was forked as the original one is fairly unmaintained."

Sounds good. But no it clearly now states it is "NO LONGER MAINTAINED".

This can be a starting point to make a decision and we should not commit this branch.

https://www.npmjs.com/package/xml-js looks a lot more popular and could do with a spike to assess the effort required to replace the existing parser.

Lets discuss.


src/modules/returns/lib/xml-adapter/mapper.js
```diff
@@ -76,7 +76,7 @@ const getMeterDetails = (ret) => {
 }
 
 const wasMeterUsed = (meterUsage) => {
-  return meterUsage.getAttribute('WasMeterUsed').value() === 'Y'
+  return meterUsage.attr('WasMeterUsed').value() === 'Y'
```
jonathangoulding, line 79: The only functional code change needed when the package was swapped. Tests still pass...

Cruikshanks (changes requested):
I don't actually want any changes. I think this is the right approach. I just don't want to merge it until we have tried deploying it to our AWS environment.

So bear with me!

jonathangoulding:
> I don't actually want any changes. I think this is the right approach. I just don't want to merge it until we have tried deploying it to our AWS environment.
> 
> So bear with me!

I will mark as draft. 

For proff of it working i have created a v22 branch - https://github.com/DEFRA/water-abstraction-service/pull/2657

Cruikshanks (approved):
🎉 

I've deployed this to our AWS `dev` environment and all seems to be well. Also, tested updating it locally and running `npm ci` in our docker-base env and all seems well there.

So it's a :shipit: from me!

Cruikshanks:
On your wider point @jonathangoulding , this is one for the business.

Essentially, our current implementation is not only unmaintained but also has vulnerabilities that will never be resolved. If we are to continue offering returns submission via XML, then we must ~update~ rebuild it in the near future.

However, knowing how little it is used, they may opt to work with the one user to help them migrate to one of our alternate methods. We can then drop the function and dependency altogether.

I'll write something up so we have a chance of this issue not being forgotten.

